PR #6, opened 2018-10-14
Fix barotropicQG forcing

navidcy:
- changes forcing implementation in `BarotropicQG` to support also deterministic time-varying forcing
- adds `test_bqg_deterministicforcingbudgets`
- fixes deprecations and syntax changes in `twodturb` and `barotropicqg` examples for `julia v1.0`


navidcy:
1. Changed `energy/enstrophy00` to `meanenergy/enstrophy`.
2. The if statement refers also to the form-stress. If there is no forcing of U(t) through `calcFU()` then the form-stress should not be computed. I'm not saying that there is not any better way to do it but I'd rather not bother now. Mostly because I am very tempted of clearing out the case with large-scale domain-averaged zonal flow U(t) from this module because it makes it complicated.
3. Agree on `using` Vs `import`; did few changes.
4. I changed `U` to a 0-dimensional array and now I was able to construct `Vars` using `structvarsexpr`.

PR #25, opened 2019-08-29
GPUify the TwoDTurb module

navidcy:
This PR modifies the `TwoDTurb` module so that it can run smoothly on both CPU and GPU.

### DIFF
--- a/src/twodturb.jl
+++ b/src/twodturb.jl
@@ -67,7 +67,7 @@ end
 Returns the params for two-dimensional turbulence.
 """
 struct Params{T} <: AbstractParams
-      ν  :: T         # Vorticity viscosity
+       ν :: T         # Vorticity viscosity
       nν :: Int       # Vorticity hyperviscous order
        μ :: T         # Bottom drag or hypoviscosity
       nμ :: Int       # Order of hypodrag
--- a/src/twodturb.jl
+++ b/src/twodturb.jl
@@ -67,11 +67,11 @@ end
 Returns the params for two-dimensional turbulence.
 """
 struct Params{T} <: AbstractParams
-  ν::T              # Vorticity viscosity
-  nν::Int           # Vorticity hyperviscous order
-  μ::T              # Bottom drag or hypoviscosity
-  nμ::Int           # Order of hypodrag
-  calcF!::Function  # Function that calculates the forcing F
+      ν  :: T         # Vorticity viscosity
+      nν :: Int       # Vorticity hyperviscous order
+       μ :: T         # Bottom drag or hypoviscosity
+      nμ :: Int       # Order of hypodrag
+  calcF! :: Function  # Function that calculates the forcing F
 end
 Params(ν, nν) = Params(ν, nν, typeof(ν)(0), 0, nothingfunction)
 
--- a/src/twodturb.jl
+++ b/src/twodturb.jl
@@ -22,11 +22,6 @@ using FourierFlows: getfieldspecs, structvarsexpr, parsevalsum, parsevalsum2
 
 abstract type TwoDTurbVars <: AbstractVars end
 
-const physicalvars = [:zeta, :u, :v]
-const transformvars = [ Symbol(var, :h) for var in physicalvars ]
-const forcedvars = [:Fh]
-const stochforcedvars = [:prevsol]
-
 nothingfunction(args...) = nothing
 
 """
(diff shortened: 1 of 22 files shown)


navidcy:
It'll be good if we can remove the boilerplate in https://github.com/FourierFlows/GeophysicalFlows.jl/blob/733833beabbc694bf3a7f49aa6948d5a77e722c1/src/twodturb.jl#L99-L127

I tried replacing it with

```julia
const physicalvars = [:zeta, :u, :v]
const transformvars = [ Symbol(var, :h) for var in physicalvars ]
const forcedvars = [:Fh]
const stochforcedvars = [:prevsol]

varspecs = cat(
  getfieldspecs(physicalvars, :(Aphys)),
  getfieldspecs(transformvars, :(Atrans)),
  dims=1)

forcedvarspecs = cat(varspecs, getfieldspecs(forcedvars, :(Atrans)), dims=1)

stochforcedvarspecs = cat(forcedvarspecs, getfieldspecs(stochforcedvars, :(Atrans)), dims=1)

struct Vars{Aphys, Atrans} <: TwoDTurbVars
  [ :( $(spec[1])::$(spec[2]) ) for spec in varspecs]
end

struct ForcedVars{Aphys, Atrans} <: TwoDTurbVars
  [ :( $(spec[1])::$(spec[2]) ) for spec in forcedvarspecs]
end

struct StochasticForcedVars{Aphys, Atrans} <: TwoDTurbVars
  [ :( $(spec[1])::$(spec[2]) ) for spec in stochforcedvarspecs]
end
```

but it didn't work... Did I do something wrong?

navidcy:
Also, when a 2D grid is created on the GPU I get this message:

```julia
┌ Warning: Performing scalar operations on GPU arrays: This is very slow, consider disallowing these operations with `allowscalar(false)`
└ @ GPUArrays ~/.julia/packages/GPUArrays/J4c3Q/src/indexing.jl:16
```

and I couldn't figure out what causes it and how to make sure the code is efficient.

src/twodturb.jl
```diff
@@ -100,50 +96,70 @@ end
...
+struct StochasticForcedVars{Aphys, Atrans} <: TwoDTurbVars
+     zeta :: Aphys
+        u :: Aphys
+        v :: Aphys
+    zetah :: Atrans
+       uh :: Atrans
+       vh :: Atrans
+       Fh :: Atrans
+  prevsol :: Atrans
+end
```
glwagner: ya!

We could maybe add special type parameters for `Fh` and `prevsol` and just have one type for all the vars. When both are `Nothing`, we have an ordinary twodturb problem. When just `prevsol` is nothing, it's a forced problem. Otherwise, its a stochastic forced problem...

glwagner: If you do that you can keep the constructor names. Just need to make sure they do the *right thing*.

test/test_twodturb.jl
```diff
@@ -172,7 +170,7 @@ function test_twodturb_advection(dt, stepper; n=128, L=2π, nu=1e-2, nnu=1, mu=0
     nothing
   end
 
-  prob = TwoDTurb.Problem(nx=n, Lx=L, nu=nu, nnu=nnu, mu=mu, nmu=nmu, dt=dt, stepper=stepper, calcF=calcF!, stochastic=false)
+  prob = TwoDTurb.Problem(nx=n, Lx=L, nu=nu, nnu=nnu, mu=mu, nmu=nmu, dt=dt, stepper=stepper, calcF=calcF!, stochastic=false, dev=dev)
```
glwagner: Should we just use greek for things like nu and mu? It's 2019 after all.

glwagner:
@navidcy I think you want to use `@eval` somehow to reduce this boiler plate. 

But there's another way --- we can define `Vars` generally, but give the optional fields special parameter types so that they can be set to `nothing` if need be. Not sure what the best way forward is.

glwagner:
> Also, when a 2D grid is created on the GPU I get this message:

Ya that's annoying. I've been getting that too in different code using `GPUifyLoops`. Your guess is as good as mine.

test/test_twodturb.jl
navidcy: hm... we've been going back and forth on this.
you know I'm prejudiced in favor of greek symbols so I'd say "yes".

glwagner: We want to be on the right side of history.

src/twodturb.jl
navidcy: Can you try do that? I've lost my patience with `@eval`...

navidcy:
> @navidcy I think you want to use `@eval` somehow to reduce this boiler plate.
> 
> But there's another way --- we can define `Vars` generally, but give the optional fields special parameter types so that they can be set to `nothing` if need be. Not sure what the best way forward is.

can you try do that? I've lost my patience with `@eval`...

navidcy:
Btw, I tried running `examples/twodturb_mcwilliams1984.jl` on GPU and it was painfully slow...
Perhaps the `diags` being CPU arrays induce constant communication back and forth from GPU to CPU?

test/test_twodturb.jl
navidcy: see eaa0f12

glwagner:
> I feel that the fact that diagnostics are saved in CPU arrays creates a bottleneck.

I don't think this by itself is a problem (you can have a CPU array of `CuArray`s, for example). But it does sound like something is awry. Diagnostics also have have to well-written to be efficient.

navidcy:
> > I feel that the fact that diagnostics are saved in CPU arrays creates a bottleneck.
> 
> I don't think this by itself is a problem (you can have a CPU array of `CuArray`s, for example). But it does sound like something is awry. Diagnostics also have have to well-written to be efficient.

OK, so regarding this PR: should we merge? Anything else needed?

PR #30, opened 2019-09-05
Fixes type instabilities in various modules

navidcy:


src/multilayerqg.jl
```diff
@@ -59,7 +59,7 @@ function Problem(;
            T = Float64)
 
    grid = TwoDGrid(nx, Lx, ny, Ly; T=T)
-   params = Params(nlayers, T(g), T(f0), T(beta), Array{T}(rho), Array{T}(H), Array{T}(U), Array{T}(eta), T(μ), T(ν), nν, grid, calcFq=calcFq)
+   params = Params(nlayers, g, f0,beta, Array{T}(rho), Array{T}(H), Array{T}(U), Array{T}(eta), μ, ν, nν, grid, calcFq=calcFq)
```
glwagner: missing space

glwagner:
Looks good! Merge when ready.

Comment: we seem to have strayed from the julia style recommendation for function arguments:

https://docs.julialang.org/en/v1/manual/style-guide/index.html#Write-functions-with-argument-ordering-similar-to-Julia-Base-1

where it is suggested that types, when used as arguments to functions should come first (think `zeros(Int, 3, 3)`, etc).

Not a big deal, but maybe something to address in the future for the `TwoDGrid` constructor.

src/multilayerqg.jl
navidcy: damn! you caught me!

PR #36, opened 2019-11-27
update TwoDTurb docs

dkarrasch:
use unicode symbols


navidcy:
thanks @dkarrasch 
actually, the docs are bit outdated and we should re-think our strategy a bit on keeping them up to date.

dkarrasch:
One strategy could be to use [https://github.com/JuliaDocs/DocStringExtensions.jl](DocStringExtensions.jl). Basically, you pepper your code with comments that would be helpful for code development anyway, and then build most of the documentation by macros. I.e., have docstrings build from the `struct` definitions, signatures of functions, fields of structs, etc. In the markdown file, one could have a few unspecific lines like "The package provides the following types/functions" and let Documenter fill in the details. Then you don't need to worry about the "hard-coded" text in the markdown, and only keep the docstrings up-to-date in the code.

PR #73, opened 2020-05-16
Plots replace PyPlot in TwoDNavierStokes examples

navidcy:
This is the beginning. We should ultimately drop PyPlot from docs/examples whatsoever.

### DIFF
--- a/examples/twodnavierstokes_decaying.jl
+++ b/examples/twodnavierstokes_decaying.jl
@@ -179,7 +179,7 @@ nothing # hide
 plot(kr, abs.(Ehr),
       linewidth = 2,
       alpha = 0.7,
-      xlabel = L"$k_r$", ylabel = L"\int \| \hat E \| k_r \mathrm{d} k_{\theta}",
+      xlabel = L"k_r", ylabel = L"\int \| \hat E \| k_r \mathrm{d} k_{\theta}",
       xlims = (5e-1, gr.nx),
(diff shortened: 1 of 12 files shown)


examples/twodnavierstokes_decaying.jl
```diff
@@ -25,13 +25,13 @@ nothing # hide
 #
 # First, we pick some numerical and physical parameters for our model.
 
-n, L  = 128, 2π             # grid resolution and domain length
+n, L  = 256, 2π             # grid resolution and domain length
```
glwagner, line 28: 😏

examples/twodnavierstokes_decaying.jl
```diff
@@ -25,13 +25,13 @@ nothing # hide
...
 nothing # hide
 
 ## Then we pick the time-stepper parameters
-    dt = 1e-2  # timestep
-nsteps = 4000  # total number of steps
- nsubs = 1000  # number of steps between each plot
+    dt = 5e-3  # timestep
+nsteps = 8000  # total number of steps
```
glwagner, line 33: This may be a pretty heavy example for docs. But we'll see how long they take to build; maybe it will be fine.

navidcy: 128 would work OK also... but 256 looks better.

The problem will arise when we make **all** examples producing 256^2 animations...

PR #75, opened 2020-05-17
Plots replace PyPlot in more examples

navidcy:


examples/barotropicqg_betaforced.jl
```diff
@@ -28,17 +28,17 @@ nx = 128       # 2D resolution = nx^2
...
 
 
 # ## Physical parameters
 
 Lx = 2π        # domain size
- ν = 0.0       # viscosity
-nν = 1         # viscosity order 
  β = 10.0      # planetary PV gradient
  μ = 0.01      # bottom drag
+ ν = 0e-5      # viscosity
```
glwagner, line 40: 0e-5?

examples/barotropicqg_betaforced.jl
```diff
@@ -69,7 +68,7 @@ seed!(1234) # reset of the random number generator for reproducibility
 nothing # hide
 
 # Next we construct function `calcF!` that computes a forcing realization every timestep
-function calcFq!(Fh, sol, t, cl, v, p, g)
+function calcFq!(Fh, sol, t, cl, vs, pr, gr)
```
glwagner, line 71: We may want to encourage a more verbose coding style in the future. But this is fine for now of course!

examples/barotropicqg_betaforced.jl
```diff
@@ -145,64 +148,81 @@ nothing # hide
...
+  pE = plot(1,
+          label="energy",
+          linewidth = 2,
+          alpha = 0.7,
+          xlims = (-0.1, 4.1),
+          ylims = (0, 0.05),
+          xlabel = "μt")
+          
+  pZ = plot(1,
+          label="enstrophy",
```
glwagner, line 213: It's of course not a big deal, but there is a typo here: there are no spaces surrounding `=` in label, while there are in the rest.

examples/barotropicqg_betaforced.jl
```diff
@@ -145,64 +148,81 @@ nothing # hide
...
-  sca(axs[6])
-  cla()
-  plot(μ*Z.t[1:Z.i], Z.data[1:E.i], label="enstrophy")
-  xlabel(L"\mu t")
-  legend()
+function plot_output(prob)
+  l = @layout grid(2, 3)
+  
+  pζ = heatmap(x, y, vs.zeta,
+            aspectratio = 1,
```
glwagner, line 155: Everything is aligned, except `aspectratio`

examples/barotropicqg_betaforced.jl
```diff
@@ -211,27 +231,36 @@ nothing # hide
...
   log = @sprintf("step: %04d, t: %d, cfl: %.2f, E: %.4f, Q: %.4f, walltime: %.2f min",
-    cl.step, cl.t, cfl, E.data[E.i], Z.data[Z.i], 
-    (time()-startwalltime)/60)
-  println(log)
+  cl.step, cl.t, cfl, E.data[E.i], Z.data[Z.i], 
+  (time()-startwalltime)/60)
+  
+  if j%(1000/nsubs)==0; println(log) end  
+  
+  p[1][1][:z] = Array(vs.zeta)
```
glwagner, line 246: I suppose the call to `Array` is to ensure this works on the GPU? Could be good to add a comment here, especially since this is an example.

examples/barotropicqg_betaforced.jl
```diff
@@ -55,7 +55,6 @@ kf, dkf = 14.0, 1.5     # forcing wavenumber and width of forcing ring in wavenu
 
 gr  = TwoDGrid(nx, Lx)
 
-x, y = gridpoints(gr)
 Kr = [ gr.kr[i] for i=1:gr.nkr, j=1:gr.nl]
 
 forcingcovariancespectrum = @. exp(-(sqrt(gr.Krsq)-kf)^2/(2*dkf^2))
```
glwagner, line 60: This line is hard to read and somewhat inscrutable. I might infer that `kf` is a forcing wavenumber... but maybe writing `forcing_wavenumber` would make this unambiguous. A comment that explains the point of this could help too. `dkf` could be named `forcing_bandwidth`.

examples/barotropicqg_betaforced.jl
```diff
@@ -55,7 +55,6 @@ kf, dkf = 14.0, 1.5     # forcing wavenumber and width of forcing ring in wavenu
 
 gr  = TwoDGrid(nx, Lx)
 
-x, y = gridpoints(gr)
 Kr = [ gr.kr[i] for i=1:gr.nkr, j=1:gr.nl]
```
glwagner, line 58: Why is this line here? I'm not sure where `Kr` is used, but the next line down seems unrelated.

examples/barotropicqg_betaforced.jl
```diff
@@ -69,7 +68,7 @@ seed!(1234) # reset of the random number generator for reproducibility
 nothing # hide
 
 # Next we construct function `calcF!` that computes a forcing realization every timestep
-function calcFq!(Fh, sol, t, cl, v, p, g)
+function calcFq!(Fh, sol, t, cl, vs, pr, gr)
   ξ = ArrayType(dev)(exp.(2π*im*rand(Float64, size(sol)))/sqrt(cl.dt))
```
glwagner, line 72: Does this allocate a new array every time-step?

You may want to allocate this array outside of the function, and then reference it as a global. I think declaring it `const` will ensure that it runs on the GPU.

Another option is to make `calcFq` a callable object that carries around a reference to `ξ` (rather than referencing it as a global). The complication with this approach is that I think you will need to use `Adapt` to get it to work on the GPU.

It's probably important to introduce what we think is best practice for this problem in the example. If its complicated, I think that means we simply need to add more simple examples to bridge the gap between simple cases and this sophisticated case.

glwagner: I guess referencing `ξ` as a global is probably a simple and adequate solution for this problem. 

I'm worried this forcing function won't work (at least not generically, or for large problems) on the GPU if it allocates every time-step, as there is no garbage collection on the GPU.

Also, the syntax on this fairly complex line is very compact, which makes it difficult to read. Putting some white space between the arithmetic operations will make this part of the code easier to read and understand.

examples/barotropicqg_betaforced.jl
navidcy: This is **exactly** the problem I wanted you to help me out.
It's a bit complicated... if I don't us `Array()` then `@animate` does not use the first frame for vorticity. Very strange....

examples/barotropicqg_betaforced.jl
navidcy: Nice catch!

navidcy: One solution is to avoid `ξ` whatsoever by:
```julia
  @. Fh = ArrayType(dev)(exp.(2π*im*rand(Float64, size(sol)))/sqrt(clock.dt))*sqrt(forcing_spectrum)
```

But that seems cumbersome, right?

navidcy: hm... I tried `@btime` and it made no difference...

navidcy:
@glwagner this PR has become huge. I'm tempted to merge as soon as tests and docs are built properly and open two issues regarding
- are `calcF!()` functions allocate arrays every time-step?
- are statements like `ξ[1, 1]=0` within `calcF!()` functions really slow down GPU? 
- why Plots.jl seems to care if, e.g., `p[1][1][:z]=prob.vars.zeta` or `p[1][1][:z]=Array(prob.vars.zeta)`?

What do you reckon?

examples/barotropicqg_betaforced.jl
glwagner: Your solution still allocates memory. Memory is allocated by this snippet:

`rand(eltype(gr), size(sol))`

Which is embedded within the expression above. This function creates an array of `size(sol)`.

This function always executes on the CPU, as well.

Note that (the last time I checked) it is not possible to generate random numbers on the GPU within a kernel (and therefore without allocation GPU memory). This function will be slow on the GPU. It may be best to write this code only to execute on the CPU, so that people are not misled into thinking it would be a good idea to run this on the GPU. I'm pretty sure this code would be very slow on the GPU.

One way to avoid allocating memory on the CPU is probably to write a function that executes a loop over the two dimensions of `ξ`. For example, you can fill an array with random values by writing

```julia
for i in eachindex(a)
    @inbounds a[i] = rand()
end
```

glwagner: It's also important to note that `ArrayType(GPU())(a)` creates a new array that has the data in `a` on the GPU, and therefore allocates memory on the GPU. Because there is no garbage collection on the GPU, the GPU will run out of memory when this function is executed many times.

glwagner: This also works:

```julia
julia> @time a .= rand(Float64, size(a));
  0.000447 seconds (9 allocations: 512.297 KiB)
```

glwagner: Here's an example:

```julia
julia> @time a .= rand(Float64, size(a));
  0.000175 seconds (9 allocations: 512.297 KiB)
```

```julia
julia> function fillrand!(a)
       for i in eachindex(a)
       @inbounds a[i] = rand()
       end
       end

julia> @time fillrand!(a)
  0.009296 seconds (6.76 k allocations: 315.136 KiB)

julia> @time fillrand!(a)
  0.000686 seconds (4 allocations: 160 bytes)
```

The function that allocates memory is actually faster in this case... but it might slow down time-stepping. I'm not sure.

glwagner: Ah, best of all:

```julia
using Random

a = zeros(256, 256)

@time rand!(a)

@time rand!(a)
```

I get 

```julia
julia> @time rand!(a);
  0.000087 seconds (4 allocations: 160 bytes)
```

navidcy:
@glwagner, since this PR is about PyPlot.jl->Plots.jl, I'm merging and moving the discussion about memory allocation and forcing functions on CPU/GPU elsewhere.

PR #77, opened 2020-05-19
[WIP] Adds temp/temph arrays

navidcy:
This addition helps clear up the code a bit and also the `calcN!()` function.

@glwagner what do you think? I can go ahead apply this to all modules...

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,22 +9,9 @@ julia:
   - 1.4
   - nightly
 
-# after_success:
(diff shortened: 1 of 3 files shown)


glwagner:
I'm not convinced this is a good idea. Consider:

1. All of the arrays in `Vars` are temporary. What is "more temporary" about this?

2. On the GPU, you're limited by the problem size you can run. Adding more arrays to `Vars` further limits the size of the problem you can do. In general its best to be as memory-lean as possible.

3. If the problem is variable naming, I would suggest removing as many elements from `Vars` as possible and replacing them all with an array `auxiliary_coeffs` and `auxiliary_grid` for Fourier coefficient arrays and grid-space arrays. Renaming / variable binding (which has negligible or zero cost for performance) in `calcN!` can be used to make the code interpretable.

What do you think?

PR #83, opened 2020-05-28
GPUify BarotropicQGQL Module

navidcy:


navidcy:
@glwagner this is the last module to be GPUified. After merging we bump up to v0.5.0.
Exciting times.

PR #84, opened 2020-05-28
Aesthetics in BarotropicQG and TwoDNavierStokes modules

navidcy:
This PR is mostly about code aesthetics. E.g., all code variables like `p, v, g, s` are now `params, vars, grid, sol` making the code much more readable.

Furthermore:
- fixes doc strings
- removes `f0` from `BarotropicQG`


navidcy:
@glwagner I'm merging since this is 99% aesthetics (with which I'm sure you'd agree).

PR #88, opened 2020-06-02
Optimize MultilayerQG on CPU

navidcy:
- `S` and `invS` are now Arrays of StaticArrays; 13x speedup on CPU


navidcy:
@glwagner can you add KernelAbstractions?

navidcy:
before these changes
```julia
julia> @btime stepforward!(prob)
  5.407 ms (133187 allocations: 9.65 MiB)
```
while after these changes:
```julia
julia> @btime stepforward!(prob)
  3.145 ms (67 allocations: 3.27 KiB)
```

glwagner:
Yes, the reason is that as it's currently written the PV inversion launches `nx * ny` very small, inefficient kernels that block GPU computation in order to do a very small matrix multiplication. This is an inefficient way to do the PV inversion.

navidcy:
@glwagner I'm thinking of merging this since it results in considerable speedup for running `MultilayerQG` on the CPU and we can explore writing kernel in a different PR. What do you reckon?

glwagner:
I think that's a very good idea. It's good to have incremental PRs anyways.

PR #103, opened 2020-09-08
Adds enstrophy budget diagnostics for 2D and barotropic QG

BrodiePearson:
I added three diagnostics to look at the enstrophy budget in anisotropic 2D turbulence (i.e. in the 2D and barotropic QG modules). The diagnostics are complementary to the existing energy budget diagnostics; the dissipation of enstrophy, the hypo-viscous/drag removal of enstrophy, and the rate of enstrophy production/destruction by the forcing.


navidcy:
thanks @BrodiePearson 
I'll have a look and let you know!

src/barotropicqg.jl
```diff
@@ -339,18 +342,30 @@ enstrophy(prob) = enstrophy(prob.sol, prob.grid, prob.vars)
...
 Returns the enstrophy of the domain-averaged U.
 """
 meanenstrophy(prob) = real(prob.params.β * prob.sol[1, 1])
 
+"""
+    dissipation_ens(prob)
+
+Returns the domain-averaged dissipation rate of enstrophy. nν must be >= 1.
+"""
+@inline function dissipation_ens(prob)
```
glwagner, line 362: a good name for this function might be `enstrophy_dissipation`

src/barotropicqg.jl
```diff
@@ -384,6 +399,25 @@ end
 
 @inline work(prob) = work(prob.sol, prob.vars, prob.grid)
 
+"""
+    work_ens(prob)
+    work_ens(sol, v, grid)
+
+Returns the domain-averaged rate of work of enstrophy by the forcing Fh.
+"""
+@inline function work_ens(sol, vars::ForcedVars, grid)
```
glwagner, line 408: A good name for this function might be `enstrophy_work`

src/barotropicqg.jl
```diff
@@ -397,4 +431,16 @@ Returns the extraction of domain-averaged energy by drag μ.
   return params.μ / (grid.Lx * grid.Ly) * parsevalsum(vars.uh, grid)
 end
 
+"""
+    drag_ens(prob)
+
+Returns the extraction of domain-averaged enstrophy by drag/hypodrag μ.
+"""
+@inline function drag_ens(prob)
```
glwagner, line 439: A good name for this function might be `drag_enstrophy_dissipation`. In light of this function, the first function could be called `viscous_enstrophy_dissipation`.

glwagner (commented):
I'll let @navidcy look at the correctness of these functions. Tests would probably be good (but can be added in a subsequent PR). I've just made some comments about naming (using full English words, and English-language patterns is best in my opinion). Great contribution!

src/barotropicqg.jl
BrodiePearson: I like your suggestions of `enstrophy_dissipation`, `enstrophy_work` and, for consistency, `enstrophy_drag`. This is consistent with the diagnostics for energetics; `dissipation`, `work` and `drag` (rather than specifying viscous dissipation and drag dissipation). 

If we took your later suggestions the energetics diagnostics should probably be changed too.

navidcy:
I agree with @glwagner function name suggestions. I think if we'll have, e.g., an `enstrophy_dissipation()` function we might as well rename the current `dissipation` to `energy_dissipation()`? 

I think that some tests would be useful. For the energy budget terms we don't test each of them separately (perhaps we should) but rather check that the budget terms add up to give the energy tendency as computed numerically, see, e.g., 

https://github.com/FourierFlows/GeophysicalFlows.jl/blob/27a13746e9dc373d0ee00b730ee269e9c8891bc0/test/test_barotropicqg.jl#L127

https://github.com/FourierFlows/GeophysicalFlows.jl/blob/27a13746e9dc373d0ee00b730ee269e9c8891bc0/test/test_barotropicqg.jl#L51

@BrodiePearson, I can help out add some enstrophy budget tests if you'd like me to. Let me know.

BrodiePearson:
@navidcy & @glwagner Thanks for the suggestions and comments! 

I changed the diagnostic names in the 2D model to be `enstrophy_dissipation`, `energy_work` etc. I also added a tests for the enstrophy budget residual, and changed the 2D forced-dissipative example to plot the enstrophy and energy budget analyses alongside each other (see below). Let me know if you see any issues! The enstrophy budget terms and the residual are a couple of orders of magnitude larger than their energy budget counterparts because there is a factor of (forcing_wavenumber)^2 difference between the enstrophy and energy injection rates. 

I did not update the BarotropicQG model any further. In another branch I have added hypo-viscosity of arbitrary power for the BarotropicQG model [in line with the 2D model capabilities], so I could bundle it in a PR for that if it would be a useful enhancement.

Budget_Diagnostics

Project.toml
```diff
@@ -10,6 +10,7 @@ FFTW = "7a1cc6ca-52ef-59f5-83cd-3a7055c09341"
 FourierFlows = "2aec4490-903f-5c70-9b11-9bed06a700e1"
 JLD2 = "033835bb-8acc-5ee8-8aae-3f567f8a3819"
 LinearAlgebra = "37e2e46d-f89d-539d-b4ee-838fcccc9c8e"
+Plots = "91a5bcdd-55d7-5caf-9e0b-520d859cae80"
```
navidcy, line 13: We don't need to add Plots as a dependency of the package. When one wants to run some of the examples they should load the `examples/Project.toml`. The docs also have Plots as a dependency; see `docs/Project.toml`.

navidcy:
@BrodiePearson this is great -- thanks!
I'll have a careful look soon.

Project.toml
BrodiePearson: Thanks, I hadn't realized those separate `Project.toml` files existed (I'm new to Julia).

navidcy: no worries!

test/test_twodnavierstokes.jl
```diff
@@ -109,28 +118,36 @@ function test_twodnavierstokes_deterministicforcingbudgets(dev::Device=CPU(); n=
...
   ii2 = (i₀+1):E.i
 
   # dEdt = W - D - R?
   total = W[ii2] - D[ii] - R[ii]
+  totalZ = WZ[ii2] - DZ[ii] - RZ[ii]
   residual = dEdt - total
+  residualZ = dZdt - totalZ
 
   isapprox(mean(abs.(residual)), 0, atol=1e-8)
+  isapprox(mean(abs.(residualZ)), 0, atol=1e-8)
```
navidcy, line 150: The way the test is written now will only return `true` or `false` for the enstrophy budget.
Last two lines of the `tests should be changed to 

```julia
return isapprox(mean(abs.(residual)), 0, atol=1e-8) & isapprox(mean(abs.(residualZ)), 0, atol=1e-8)
```

test/test_twodnavierstokes.jl
```diff
@@ -56,30 +56,39 @@ function test_twodnavierstokes_stochasticforcingbudgets(dev::Device=CPU(); n=256
...
   # total = W[ii2]+ε - D[ii] - R[ii]      # Ito
   total = W[ii2] - D[ii] - R[ii]        # Stratonovich
+  totalZ = WZ[ii2] - DZ[ii] - RZ[ii]
 
   residual = dEdt - total
   isapprox(mean(abs.(residual)), 0, atol=1e-4)
+
+  residualZ = dZdt - totalZ
+  isapprox(mean(abs.(residualZ)), 0, atol=(kf^2)*1e-4)
 end
```
navidcy, line 92: Similarly, this needs to be changed to:

```julia
 residual = dEdt - total
residualZ = dZdt - totalZ

return isapprox(mean(abs.(residual)), 0, atol=1e-4) && isapprox(mean(abs.(residualZ)), 0, atol=(kf^2)*1e-4)
```

test/test_twodnavierstokes.jl
```diff
@@ -56,30 +56,39 @@ function test_twodnavierstokes_stochasticforcingbudgets(dev::Device=CPU(); n=256
...
   # dEdt = W - D - R?
   # If the Ito interpretation was used for the work
   # then we need to add the drift term
   # total = W[ii2]+ε - D[ii] - R[ii]      # Ito
   total = W[ii2] - D[ii] - R[ii]        # Stratonovich
+  totalZ = WZ[ii2] - DZ[ii] - RZ[ii]
 
   residual = dEdt - total
   isapprox(mean(abs.(residual)), 0, atol=1e-4)
+
```
navidcy, line 89: This needs to be changed to:

```julia
 residual = dEdt - total
residualZ = dZdt - totalZ

return isapprox(mean(abs.(residual)), 0, atol=1e-4) && isapprox(mean(abs.(residualZ)), 0, atol=(kf^2)*1e-4)
```
otherwise only the answer regarding the enstrophy budget is taken into consideration.

navidcy:
I'm not ignoring this @BrodiePearson. While reviewing this I realised that there was a bug in the tests for deterministic forcing energy budgets for `TwoDNavierStokes`. I fixed it in #105.

examples/twodnavierstokes_stochasticforcing.jl
```diff
@@ -125,23 +130,28 @@ nothing # hide
...
   t = E.t[ii]
   dEdt_computed = W[ii2] - D[ii] - R[ii]        # Stratonovich interpretation
-  
-  residual = dEdt_computed - dEdt_numerical
+  dZdt_computed = WZ[ii2] - DZ[ii] - RZ[ii]
+
+  residual_E = dEdt_computed - dEdt_numerical
+  residual_Z = dZdt_computed - dZdt_numerical
+
+  εZ = ε*(forcing_wavenumber^2)  # Estimate of mean enstrophy injection rate
```
navidcy, line 150: This is correct since the forcing is "narrow-ish" in wavenumber space.
A better way is:
```julia
εZ= parsevalsum(forcing_spectrum/2, g)/(g.Lx * g.Ly)
```
right?

navidcy (changes requested):
I suggest splitting the `test_twodnavierstokes_stochasticforcingbudgets()` to `test_twodnavierstokes_stochasticforcing_energybudgets()` and `test_twodnavierstokes_stochasticforcing_enstrophybudgets()`. Similarly for `...deterministicforcing_...`. But this can be also done after we merge!

This is a nice addition. Thanks @BrodiePearson. I suggest we merge and we can add some tests for the enstrophy budgets in `BarotropicQG` later.

Regarding the hypoviscosity in `BarotropicQG`, I have never seen anybody using hypo-viscosity in a QG setup. Do you have anything in mind? I was under the impression that hypo-viscosity was only a "trick" to dissipate energy piling up at largest scale without affecting the inertial range and that was only used in 2D turbulence simulations.

.gitignore
```diff
@@ -19,6 +19,7 @@ docs/Manifest.toml
 # Model output
 **/*.jld
 **/*.jld2
+**/*.mat
```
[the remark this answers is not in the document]
BrodiePearson: I've been post-processing with some Matlab functions so I've been outputting .mat in addition to .jld* files. I expect you won't be having any .mat-producing examples in the master you can remove it if you like and I'll just keep it on my fork.

examples/twodnavierstokes_stochasticforcing.jl
BrodiePearson: Good suggestion! Changed it

BrodiePearson:
@navidcy Thanks for catching those issues and for suggesting resolutions. I made all the suggested changes and also fixed some typos in the example figures. Feel free to make any modifications if you catch more.

As for the hypo-drag, I'm using the barotropic QG model to simulate anisotropic 2D turbulence (i.e. 2D turbulence on a sphere). I'm interested in the statistics of the inverse energy and direct enstrophy cascades as a function of the flow anisotropy, and hypo-drag gets used for these types of simulation (e.g. [this study](https://aip.scitation.org/doi/10.1063/1.1327594)). I'm planning to use the 2D and BarotropicQG models together to achieve this, and since hypo-drag was critical to get a nice isotropic inverse cascade in the 2D model I wanted to be able to run exactly comparable simulations in BarotropicQG.

I'm hoping this will complement an analogous study under a range of QG conditions using the other GeophysicalFlows models. If you'd like I would be happy to discuss my plans in more detail, I think GeophysicalFlows is going to be central to my modelling efforts.

Issue #90, opened 2020-06-03
Animate the TwoDNavierStokes forced-dissipative example


navidcy:
@thdrivas urges us to add a movie showing the evolution of the TwoDNavierStokes forced-dissipative example:

https://fourierflows.github.io/GeophysicalFlowsDocumentation/dev/generated/twodnavierstokes_stochasticforcing/

PR #108, opened 2020-09-11
Separate 2D Navier-Stokes stochastic forcing & budgets examples

navidcy:
- separates 2D Navier-Stokes forced-dissipative example into
  (i) an example that plots a movie of the vorticity field + evolution of energy/entstrophy time series
  (ii) an example that plots the terms of the energy and enstrophy budgets

- adds some comments/clarifications in examples

@BrodiePearson could you have a quick look at this PR?

As soon as the Documentation is built, you should be able to see a preview of how the docs should look like after merging at
https://fourierflows.github.io/GeophysicalFlowsDocumentation/previews/PR108/


navidcy:
This PR closes #90.

examples/barotropicqg_betaforced.jl
```diff
@@ -53,26 +53,27 @@ forcing_wavenumber = 14.0    # the central forcing wavenumber for a spectrum tha
 forcing_bandwidth  = 1.5     # the width of the forcing spectrum 
 ε = 0.001                    # energy input rate by the forcing
 
-gr  = TwoDGrid(nx, Lx)
+gr  = TwoDGrid(n, L)
```
BrodiePearson, line 56: Is this necessary? I noticed the betadecay example doesn't explicitly define the grid, so is there a reason the betaforced example does?

BrodiePearson (approved):

examples/barotropicqg_betaforced.jl
navidcy: Because wave numbers are needed to construct the forcing spectrum below. I know it’s not neat but I can’t think of a better way.

BrodiePearson: Ah I see. The decay example pulls the grid from the initialized problem, but the forced example needs the grid in order to initialize the problem with that spectrum. Thanks!

PR #111, opened 2020-09-15
Adds Surface QG model and example

BrodiePearson:
This PR adds a surface quasi-geostrophic (SQG) module `surfaceqg.jl,` some basic documentation for the module, and an example (`surfaceqg_decaying.jl`) that reproduces the spin-down of an initial elliptic buoyancy field as simulated by [Held et al., 1995](http://pordlabs.ucsd.edu/wryoung/theorySeminar/pdf14/HeldPierrehumbert.pdf). The SQG module solves the evolution equation for surface buoyancy, in contrast to the (potential) vorticity evolution of other modules. The relation between buoyancy, streamfunction, and velocities are summarized in the documentation.

Below are snapshots of the buoyancy field from this module's example and Figure 2 of their paper. They agree very well! This example is also used by [another SQG model](https://pyqg.readthedocs.io/en/latest/examples/sqg.html), allowing further comparison/validation.

Some preliminary diagnostics have been coded into the module, but a test of the budgets will be included in a later PR. A preliminary test suggests that the budgets do not close - so there may be some mistakes in the current diagnostic definitions (they are not used in the above example).

image


BrodiePearson:
@navidcy I am not sure how to view the documentation associated with this PR, so I can't test whether the new .md file and example render successfully in the documentation.

navidcy:
@BrodiePearson, thanks for the PR!

Regarding the docs, this line
https://github.com/BrodiePearson/GeophysicalFlows.jl/blob/29085e50aa2b128badb1b5e255f88b9b50b31b93/docs/make.jl#L103
results in PR pushing their docs in a `previews` folder.
The preview of the Docs should have appeared at 
https://fourierflows.github.io/GeophysicalFlowsDocumentation/previews/PR111/

Looking at the Documentation build log I think you are missing some GitHub secret keys and that's why there was nothing pushed in the Docs...

```julia
Load key "/home/runner/work/GeophysicalFlows.jl/GeophysicalFlows.jl/docs/.documenter": invalid format
git@github.com: Permission denied (publickey).
fatal: Could not read from remote repository.

Please make sure you have the correct access rights
and the repository exists.
┌ Error: Git failed to fetch git@github.com:FourierFlows/GeophysicalFlowsDocumentation.git
│ This can be caused by a DOCUMENTER_KEY variable that is not correctly set up.
│ Make sure that the environment variable is properly set up as a Base64-encoded string
│ of the SSH private key. You may need to re-generate the keys with DocumenterTools.
```

Let me look into it and get back to you.

Until then you can build the Docs locally and see how they look! From the repo's home directory try calling:
```
$ julia --project=docs/ -e 'using Pkg; Pkg.instantiate(); Pkg.develop(PackageSpec(path=pwd()))'; julia --project=docs/ docs/make.jl; open docs/build/index.html
```

glwagner:
Wow, pretty sweet vortex!

src/surfaceqg.jl
```diff
@@ -0,0 +1,370 @@
...
+
+"""
+    calcN_advection(N, sol, t, clock, vars, params, grid)
+
+Calculates the advection term.
+"""
+function calcN_advection!(N, sol, t, clock, vars, params, grid)
+  @. vars.bh = sol
+  @. vars.uh =   im * grid.l  * sqrt.(grid.invKrsq) * sol
+  @. vars.vh = - im * grid.kr * sqrt.(grid.invKrsq) * sol
```
glwagner, line 169: Don't need the extra `.` in `sqrt` --- the `@.` macro in front of this expression will put dots in front of every function. So this line can read

```julia
@. vars.vh = - im * grid.kr * sqrt(grid.invKrsq) * sol
```

The same applies to the line above.

This is not a huge deal of course, since it works as-is.

src/surfaceqg.jl
```diff
@@ -0,0 +1,370 @@
...
+  return 1 / (grid.Lx * grid.Ly) * parsevalsum(vars.uh, grid)
+end
+
+"""
+    buoyancy_variance(prob)
+
+Returns the domain-averaged buoyancy variance. In SQG flows this is identical to
+the domain-averaged velocity variance (twice the kinetic energy)
+"""
+@inline function buoyancy_variance(prob)
```
glwagner, line 264: 👏🏼

src/surfaceqg.jl
```diff
@@ -0,0 +1,370 @@
...
+
+"""
+    buoyancy_work(prob)
+    buoyancy_work(sol, v, grid)
+
+Returns the domain-averaged rate of work of buoyancy variance by the forcing Fh.
+"""
+@inline function buoyancy_work(prob)
+  sol, vars, params, grid = prob.sol, prob.vars, prob.params, prob.grid
+  @. vars.uh =  sol * conj(vars.Fh)
```
glwagner, line 292: A comment here about how `uh` is acting as a temporary variable for... something... (the Fourier transform of the rate of work of buoyancy variance?) could be helpful.

src/surfaceqg.jl
```diff
@@ -0,0 +1,370 @@
...
+  return 1 / (grid.Lx * grid.Ly) * parsevalsum(vars.uh, grid)
+end
+
+@inline function buoyancy_work(sol, vars::StochasticForcedVars, grid)
+  @. vars.uh =  (vars.prevsol + sol) / 2 * conj(vars.Fh) # Stratonovich
+  # @. vars.uh = grid.invKrsq * vars.prevsol * conj(vars.Fh)           # Ito
```
glwagner, line 298: I supposed this commented-out line appears in the `TwoDimTurbulence` module. But I find it confusing --- which is right? Should this line be uncommented sometimes? Removing it may be the best course if we have decided that "Stratonovich" is either correct, or the convention that we are using when calculating budgets. 

@navidcy any comments?

glwagner: Another option is to expand the written comments here to explain when this line should be uncommented.

src/surfaceqg.jl
```diff
@@ -0,0 +1,370 @@
...
+
+Returns the average of domain-averaged advection of kinetic energy by the
+leading-order (geostrophic) flow.
+"""
+@inline function kinetic_energy_advection(prob)
+  sol, vars, params, grid = prob.sol, prob.vars, prob.params, prob.grid
+
+  @. vars.b = vars.u * vars.u
+  mul!(vars.bh, grid.rfftplan, vars.u)
+  sol = vars.bh
```
glwagner, line 355: This line re-binds the name `sol` to `vars.bh`. It does not copy the data in `bh` into the memory space occupied by `prob.sol`.

src/surfaceqg.jl
```diff
@@ -0,0 +1,370 @@
...
+  @. vars.v = vars.v * vars.u
+  mul!(vars.bh, grid.rfftplan, vars.u)
+  @. vars.uh = - ( im * grid.kr * sol + im * grid.l * vars.bh ) * conj(vars.uh)
+
+  @. vars.u = vars.u * vars.v
```
glwagner, line 360: Can be written

```julia
vars.u .*= vars.v
```

src/surfaceqg.jl
```diff
@@ -0,0 +1,370 @@
...
+    kinetic_energy_advection(prob)
+
+Returns the average of domain-averaged advection of kinetic energy by the
+leading-order (geostrophic) flow.
+"""
+@inline function kinetic_energy_advection(prob)
+  sol, vars, params, grid = prob.sol, prob.vars, prob.params, prob.grid
+
+  @. vars.b = vars.u * vars.u
+  mul!(vars.bh, grid.rfftplan, vars.u)
```
glwagner, line 354: Is the third argument supposed to be `vars.b`?

src/surfaceqg.jl
```diff
@@ -0,0 +1,370 @@
...
+  sol = vars.bh
+  @. vars.v = vars.v * vars.u
+  mul!(vars.bh, grid.rfftplan, vars.u)
+  @. vars.uh = - ( im * grid.kr * sol + im * grid.l * vars.bh ) * conj(vars.uh)
+
+  @. vars.u = vars.u * vars.v
+  mul!(vars.bh, grid.rfftplan, vars.u)
+  sol = vars.bh
+  @. vars.v = vars.v * vars.v
```
glwagner, line 363:
```suggestion
  vars.v .*= vars.v
```

src/surfaceqg.jl
navidcy: I think we should remove all Ito/Stratonovich remarks and perhaps expand the [docs section](https://fourierflows.github.io/GeophysicalFlowsDocumentation/dev/forcing/) or write up a literated example on this manner.

src/surfaceqg.jl
```diff
@@ -0,0 +1,370 @@
...
+  mul!(vars.bh, grid.rfftplan, vars.v)
+  @. vars.vh = - ( im * grid.kr * sol + im * grid.l * vars.bh ) * conj(vars.vh)
```
glwagner, line 365: What's going on here? It looks like there's some finagling with temporary variables that leads to possibly confusing expressions. Comments could help. Another option is to use different names. For example we could use something like

```julia
u² = vars.b
@. u² = vars.u^2

vars.v .*= vars.u
uv = vars.v
```

et cetera.

are perfectly fine.

navidcy: Doesn't this induce allocating new arrays in memory @glwagner ?

glwagner (commented):
Awesome to have a new module, especially one that can make pretty pictures! :-D

I have left a few minor comments in the code which hopefully will include its clarity. I also think it could be a good idea to include some simple tests in this PR that ensure, for example, that `SurfaceQG.Problem` can be used to construct a problem without throwing an error, or that `stepforward!(prob)` when `prob` comes from `SurfaceQG` can be called without error (and perhaps the same for the other user-facing functions this PR).

With unit tests in place, I think its perfectly ok if more involved / quantitative dynamics and budgets tests come in a future PR --- though it might be good to add a remark to this effect in the docs (since future PRs are never certain!)

src/surfaceqg.jl
navidcy: I think that's OK but 2 lines **below** should be ` mul!(vars.bh, grid.rfftplan, vars.v)`?

navidcy:
I can help adding tests (but you'll have to give me few days, e.g., during weekend). ;)

src/surfaceqg.jl
glwagner: No.

```julia
julia> a = rand(1024, 1024);

julia> @time for i = 1:100
       b = a
       end
  0.000000 seconds

julia> @time for i = 1:100
       c = deepcopy(a)
       end
  0.270146 seconds (400 allocations: 800.043 MiB, 39.13% gc time)
```

The equals sign (`=`) performs a name binding --- in other words `b=a` when `a` is an array means "bind the name `b` to the memory in `a`). Then `b[1, 1]` accesses the _same_ memory as `a[1, 1]`. `b=a` does not copy an array. For that, you must use `copy` or `deepcopy`.

glwagner: If there is an _expression_ on the right side of the equals sign that requires memory allocation, then a new array will be created for that purpose, and the memory associated with that new array will be bounded to the name on the left side of the equal sign. This is what occurs when you write `b = a .* c`.

glwagner: The classic demonstration of my first comment is

```julia
julia> a = rand(2, 2)
2×2 Array{Float64,2}:
 0.442335  0.153258
 0.377039  0.558756

julia> b = a
2×2 Array{Float64,2}:
 0.442335  0.153258
 0.377039  0.558756

julia> a[1, 1] = 0
0

julia> b
2×2 Array{Float64,2}:
 0.0       0.153258
 0.377039  0.558756
```

navidcy: dang! of course..

(these sort of name-bindings can help clarify code in other modules btw..)

navidcy:
pushed directly to @BrodiePearson's master -- sorry!
I should have pushed on a different branch on @BrodiePearson's fork and ask for a PR.

navidcy:
Hm, what's the best workflow here... pushing PR's to @BrodiePearson's fork could become tedious...
Perhaps @BrodiePearson should have opened a PR from `BrodiePearson:master` -> `FourierFlows:sqg` and then we all commit there?

navidcy:
> Until then you can build the Docs locally and see how they look! From the repo's home directory try calling:
> 
> ```
> $ julia --project=docs/ -e 'using Pkg; Pkg.instantiate(); Pkg.develop(PackageSpec(path=pwd()))'; julia --project=docs/ docs/make.jl; open docs/build/index.html
> ```

btw, @BrodiePearson building/literating all examples could take a while so when I'm doing that usually I comment out all docs modules except the one I'm developing. Remember to delete # before committing ;)

src/surfaceqg.jl
```diff
@@ -0,0 +1,370 @@
...
+# Equations
+# ---------
+
+"""
+    Equation(params, grid)
+
+Returns the equation for Surface QG turbulence with params p and `grid`.
+"""
+function Equation(params::Params, grid::AbstractGrid)
+  L = @. - params.ν * grid.Krsq^params.nν - params.μ * grid.Krsq^params.nμ
```
navidcy, line 92: This includes hypoviscosity but the docs don't mention anything. Shall we keep hypoviscosity here and add it in the equation in the docs or remove it from here whatsoever?

BrodiePearson: I don't think I have seen an SQG model with hypo-viscosity since energy and buoyancy variance cascades to small scales. I would support removing it.

BrodiePearson:
@navidcy Would you like me to resubmit a PR to a separate branch as you suggested above, or keep working on this here?

navidcy:
Let's keep working here. Is it OK if I commit to your master or I'll mess up things?
(I already committed few times but it was all SQG related so hopefully that didn't do any harm...)

navidcy:
Perhaps you can address @glwagner 's comments and then I'll work on tests?

BrodiePearson:
Sounds good! Feel free to commit to my master. It's only SQG stuff on my master at the moment

navidcy:
@BrodiePearson docs mention `surfaceqg_decaying_budget.jl`. Will you push this example or shall I remove the line in docs?

src/surfaceqg.jl
BrodiePearson: These lines have gone as they were unnecessary (already had the Fourier transformed variables to work with), but I made this mistake a lot through the file so I've corrected all those instances.

src/surfaceqg.jl
BrodiePearson: I added a comment about what that is. This could be made even clearer by adding equations to the text above each function that appears in the documentation)

src/surfaceqg.jl
BrodiePearson: There were a bunch of errors in the advection diagnostics. I think I fixed them and I added some comments to describe what was going on in each step. I tried these diagnostics in the example and it was encouraging; buoyancy variance advection is very small (conserved!), but the kinetic energy advection is large (non-conserved due to ageostrophic effects)

I had to bring sol in because the advection operator is the sum of several terms. I needed a scratch variable that was the right size to be multiplied by some function of velocities (vars.uh and vars.vh) and then added to the advection (stored as vars.bh)  as each term was calculated. If you can think of a better way to do this I'm all ears!

BrodiePearson:
@navidcy You can remove that line, I plan to have something pretty soon but I need to think a little more about the kinetic energy diagnostics.

navidcy:
I think if we add few tests and confirm that the example still works then it's ready!

BrodiePearson:
Great! I am going to head to sleep now but I'll try and make some conceptual headway on the budgets tomorrow. Thanks for your help in getting this PR to a mergeable state!

src/surfaceqg.jl
glwagner: Well, just two concerns: the line this comment refers to does not copy the data in `vars.bh` to `sol`. That would require `sol .= vars.bh` (note the `.` before the `=`). Instead, this line creates a new "name" called `sol` (therefore discarding the original reference to `sol`, which is an alias for `vars.bh`. I suspect this is not what was intended.

The second issue is that modifying `sol` causes the model state to be lost, since `sol` keeps track of the model state (the state of the model is not kept anywhere else).

Two alternative solutions may be used: 1) add a new temporary variable to `vars` to be used for this diagnostic. 2) I am not certain, but because the object being calculated has two terms, I think that three temporary variables may be adequate. One strategy you could use is to store the first term in this diagnostic in `vh` (for example), and then use the remaining two temporary variables (possibly overwriting their previous data) to compute the second term, adding that contribution to the already calculated component stored in `vh`. Hope that makes sense.

navidcy: I fully agree; `sol` should not be used as scratch variable.
@BrodiePearson, how about this:

```Julia
 @inline function kinetic_energy_advection(prob)
  sol, vars, params, grid = prob.sol, prob.vars, prob.params, prob.grid

  updatevars!(prob)

  mul!(vars.vh, grid.rfftplan, vars.u.^2)
  @. vars.bh = - im * grid.kr * vars.vh * conj(vars.uh)  # - FFT[∂(u²)/∂x]*conj(uh)
  mul!(vars.vh, grid.rfftplan, vars.u .* vars.v)
  @. vars.bh -= im * grid.l * vars.vh * conj(vars.uh)    # - FFT[∂(uv)/∂y]*conj(uh)

  @. vars.vh = - im * grid.kr * sqrt(grid.invKrsq) * sol # because vars.vh was used as scratch

  mul!(vars.uh, grid.rfftplan, vars.u .* vars.v)
  @. vars.bh -= im * grid.kr * vars.uh * conj(vars.vh)    # - FFT[∂(uv)/∂x]*conj(vh)
  mul!(vars.uh, grid.rfftplan, vars.v.^2)
  @. vars.bh -= im * grid.l * vars.uh * conj(vars.vh)     # - FFT[∂(v²)/∂y]*conj(uh)

  return 1 / (grid.Lx * grid.Ly) * parsevalsum(vars.bh , grid)
end
```

I'm not sure if these are the terms you want to compute; I tried to do what the comments on the `kinetic_energy_advection(prob)` function were suggesting.

Notice how I call `updatevars!(prob)` in the beginning to make sure my variables have what they supposed to have. Also, I need to update `vars.vh` after the first half of the computation is done?

src/surfaceqg.jl
```diff
@@ -0,0 +1,363 @@
...
+  mul!(vars.uh, grid.rfftplan, vars.u .* vars.v) # Transform adv. of u * v
+  # add -FFT(∂[uv]/∂x) * conj(FFT(v))
+  @. vars.bh -= im * grid.kr * vars.uh * conj(vars.vh)
+  mul!(vars.uh, grid.rfftplan, vars.v .* vars.v) # Transform adv. of v * v
+  # add -FFT(∂[vv]/∂y) * conj(FFT(v))
+  @. vars.bh -= im * grid.l * vars.uh * conj(vars.vh)
+  # vars.bh is now -conj(FFT[u̲])⋅FFT[u̲⋅∇(u̲)]
+
+  return 1 / (grid.Lx * grid.Ly) * parsevalsum( vars.bh , grid)
+end
```
BrodiePearson, line 361: I think I have the advection terms right in this commit. I changed it so that sol is no longer modified in the function - and velocities/buoyancy are derived from sol in case their vars.vh etc. values are not in sync with the solution. I tried to comment parts of code where variables being used as temporary storage, and parts where they are the true variable. @glwagner I agree your suggestion about new variable names could be really helpful here and in other diagnostics, but I wasn't comfortable enough with the method to make those changes here.

navidcy:
@BrodiePearson, the kinetic energy advection isn't it identically zero when integrated over the domain?
Isn't kinetic_energy_advection supposedly computing the domain average of 
u ∂( KE )/∂x + v ∂( KE )/∂y
?

BrodiePearson:
@navidcy The buoyancy is only advected and dissipated in this system, so the buoyancy advection should average to zero (I found it was a considerable sink of buoyancy variance in some preliminary tests - perhaps that is an issue with the advection scheme in SQG dynamics).

The kinetic energy is not conserved. The budget has advection, viscosity, and an ageostrophic term due to the vertical gradient of vertical velocity  [Equation 2.12 of this paper](http://gershwin.ens.fr/glapeyre/papers/xcpklhgljmcw08.pdf). It would be great to add that as a diagnostic [using Eqs. 2.7 and 2.8] and validate that Eq. 2.12 is the budget seen by the numerical scheme.

navidcy:
Equation 2.12 in Capet 2012 refers to each individual Fourier components of the KE. But if `kinetic_energy_advection` is supposed to be computing u ∂( KE )/∂x + v ∂( KE )/∂y (modulo in Fourier space) and **integrate it over the domain** then it must be identically zero, right? From what I understand, if one wants to compute the budget for each Fourier component then advection terms matter but if you are interested in the domain-integrated budget they shouldn't.

I've added tests for everything **but** the `buoyancy_advection` and `kinetic_energy_advection` functions. I suggest we merge the PR as is now **without** the `buoyancy_advection` and `kinetic_energy_advection` functions and then open a *new* PR to add `buoyancy_advection` and `kinetic_energy_advection` there and continue the discussion there. What do you think? I can do that if you agree.

navidcy:
> (I found it was a considerable sink of buoyancy variance in some preliminary tests - perhaps that is an issue with the advection scheme in SQG dynamics or the fact it was using an early code version).

Perhaps it was related with variables used as scratch and not having the arrays functions were expecting them to have...

navidcy:
I found also a typo; the buoyancy_work should have been multiplied with a 2 because buoyancy variance is defined as B = ∫ b² d²x (without 1/2 factor, like the enstrophy) and, therefore, dB/dt = ∫ 2 b ∂b/∂t d²x = ∫ 2 b f d²x  + ...

BrodiePearson:
That all sounds good! I agree with you about the integrated advection terms being zero, so it makes sense to take those diagnostics out for this PR (I'm not even convinced we should add them back in a later PR).

navidcy (approved):

navidcy:
OK @BrodiePearson I'm merging -- thanks again.
I'll let you decide on whether a PR for adding the `buoyancy_advection()` and `kinetic_energy_advection()` functions is in order.

PR #145, opened 2020-11-23
Enhance examples + Fix some plot orientations

### LINKED ISSUE #134: Some of the animations have the axes flipped
Back in the day when I was learning Plots.jl I didn't understand that I needed to transpose the array for proper orientation.

E.g.,
https://github.com/FourierFlows/GeophysicalFlows.jl/blob/81048e1da0948fb4cc02b271d0bfd1afc8812fa3/examples/twodnavierstokes_decaying.jl#L61
should read
```
heatmap(x, y, vs.zeta', 
```

and this has propagated into the examples (especially when we plot fields that are isotropic in both directions.

navidcy:
This PR just goes through the example once again. 

- Μakes sure that fields are plotted with the correct orientation, e, g., `heatmap(x, y, ζ)` -> `heatmap(x, y, ζ')`.
- Changes `$math$` to ` ``math`` ` for inline latex expressions.
- Adds `cfl` in all example logs and make sure cfl is only computed whenever log is printed out.
- Makes expressions more verbose, e.g., `p, g, v, cl` ->`params, grid, vars, clock`.
- Beautifies code where possible.

A preview is found here: https://fourierflows.github.io/GeophysicalFlowsDocumentation/previews/PR145

Closes #134.

### DIFF
--- a/examples/twodnavierstokes_stochasticforcing.jl
+++ b/examples/twodnavierstokes_stochasticforcing.jl
@@ -44,16 +44,16 @@ nothing # hide
 # width ``\delta k_f``, and it injects energy per unit area and per unit time equal
 # to ``\varepsilon``.
 
-forcing_wavenumber = 14.0    # the central forcing wavenumber for a spectrum that is a ring in wavenumber space
-forcing_bandwidth  = 1.5     # the width of the forcing spectrum
-ε = 0.1                      # energy input rate by the forcing
+forcing_wavenumber = 14.0 * 2π/L   # the central forcing wavenumber for a spectrum that is a ring in wavenumber space
+forcing_bandwidth  = 1.5  * 2π/L   # the width of the forcing spectrum
+ε = 0.1                            # energy input rate by the forcing
 
 grid = TwoDGrid(dev, n, L)
 
 K = @. sqrt(grid.Krsq)
 forcing_spectrum = @. exp(-(K - forcing_wavenumber)^2 / (2 * forcing_bandwidth^2))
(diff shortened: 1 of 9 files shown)


examples/twodnavierstokes_stochasticforcing.jl
```diff
@@ -52,10 +52,10 @@ grid = TwoDGrid(dev, n, L)
 
 K = @. sqrt(grid.Krsq)
 forcing_spectrum = @. exp(-(K - forcing_wavenumber)^2 / (2 * forcing_bandwidth^2))
-forcing_spectrum[K .< ( 2 * 2π/L)] .= 0 # no power at low wavenumbers
-forcing_spectrum[K .> (20 * 2π/L)] .= 0 # no power at high wavenumbers
+@. forcing_spectrum = ifelse(K < 2, 0, forcing_spectrum)       # no power at low wavenumbers
```
BrodiePearson, line 55: I see now (after reviewing your other PR) that you're removing the 2 pi/L factors from all the forcing terms in the examples. I guess that makes sense in terms of streamlining the code, since most of these domains have width 2 pi, but do you think it might confuse people if they adapt the domain size for their own simulations.

BrodiePearson (commented):
All the changes look good to me.

BrodiePearson (approved):

examples/twodnavierstokes_stochasticforcing.jl
navidcy: That was accidental. I'm putting the `L/2π` factors back -- thanks for noticing!

PR #148, opened 2020-11-27
[WIP] Split updatevars!() for Fourier transformed variables

### LINKED ISSUE #81: Separate updatevars! for physical space and Fourier space
Updating variables in physical space requires an inverse Fourier transform, which in turns requires a `deepcopy()`. Sometimes, though, we only need to update the Fourier transform of variables. Therefore if we split `updatevars!` into two separate function we might gain on allocation.

navidcy:
Closes #81

### DIFF
--- a/src/twodnavierstokes.jl
+++ b/src/twodnavierstokes.jl
@@ -214,18 +214,34 @@ end
 # ----------------
 
 """
+    updatetransfomedvars!(prob)
+
+Update the Fourier transformed variables in `vars` with solution in `sol`.
+"""
+function updatetransfomedvars!(prob)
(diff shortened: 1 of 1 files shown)


navidcy:
@glwagner, I reckon you were thinking something like this, right?

glwagner (approved):
When was I thinking of something like this?

navidcy:
haha, we were chatting about it over txt or call (I don't remember).

https://github.com/FourierFlows/GeophysicalFlows.jl/issues/81

navidcy:
But now that I look closer it only makes sense for MultilayerQG and in particular for the fluxes diagnostics; I'l revise.

PR #157, opened 2020-12-02
Convert all animations in Docs from mp4 to gifs

### LINKED ISSUE #113: Animations in Docs don't show up in Safari

navidcy:
https://fourierflows.github.io/GeophysicalFlowsDocumentation/previews/PR157/

Closes #113

### DIFF
--- a/docs/make.jl
+++ b/docs/make.jl
@@ -51,7 +51,7 @@ Timer(t -> println(" "), 0, interval=240)
 format = Documenter.HTML(
   collapselevel = 2,
      prettyurls = get(ENV, "CI", nothing) == "true",
-      canonical = "https://fourierflows.github.io/GeophysicalFlowsDocumentation/dev/"
+      canonical = "https://fourierflows.github.io/GeophysicalFlowsDocumentation/stable/"
 )
 
 makedocs(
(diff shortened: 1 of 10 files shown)


glwagner (approved):
Love gif's! I usually use fps=8 but if you have a lot of output it makes sense to speed it up...

navidcy:
Do you reckon the animations go fast? They look good on my screen.

navidcy:
(The only problem with gifs is that they are bigger in file size...)

glwagner:
I think you could decrease the number of outputs and change to fps=8 to speed up the runs, that's all.

glwagner:
Right, so decreasing the number of snapshots and also slowing the movies down to obtain a similar effect would also reduce the file size.

navidcy:
Well, the numbers here are the minimum that my eyes and brain feel good with. E.g., examples with stochastic forcing become too flickery...

But if you wanna have a go trying to reduce the numbers go for it :)

navidcy:
Damn... you got me started now :)

glwagner:
I think it's ok if the movies don't look "continuous". The question is not "how pretty" but "how useful", I think.

navidcy:
as pretty as one can get! :)

navidcy:
I'll merge now and we can work on this at some later time. Until then I'll try to keep cleaning the GeophysicalFlowsDocumentation repo.

PR #158, opened 2020-12-03
Def radius

liasiegelman:


navidcy:
hey, @liasiegelman let's chat... perhaps we can clean it a bit?
seems like you submitted 2 .jl files
ideally you just want to take the `barotropicqg.jl`, rename it and modify it.
this way you'll do the least modification to the code.

liasiegelman:
@navidcy ah yes that was unintentional, just wanted to submit SingleLayerQG. I have a meeting at 2pm. Could we chat later in the afternoon your time? Meanwhile, could you reject my pull request and I'll try to submit a cleaner one? Thanks !

PR #160, opened 2020-12-04
SingleLayerQG

### LINKED ISSUE #137: Add 1.5-layer QG functionality
It's easy to include a finite Rossby deformation radius in BarotropicQG module. I vote first to clear up `BarotropicQG` module as implied by #135 and then modify the BarotropicQG module to add such functionality.

pinging @liasiegelman as she's interested and could help us out here.

liasiegelman:
This PR renames the BarotropicQG module to SingleLayerQG and adds the ability to include finite Rossby radius of deformation. Also takes the chance to refactor/optimize certain features.

Closes #137.


navidcy:
Great! That's a good start.

What we need to make sure we have is:

- [ ] deformation radius feature + kinetic energy / potential energy functions
- [ ] tests for the new features
- [ ] make sure all previous BarotropicQG examples run + add a SingleLayerQG example with non-zero deformation wavenumber
- [ ] update the Docs

src/singlelayerqg.jl
```diff
@@ -183,7 +187,8 @@ end
 
 function calcN_advection!(N, sol, t, clock, vars, params, grid)
   @. vars.zetah = sol
-  @. vars.psih  = -grid.invKrsq * vars.zetah
+  @. vars.psih  = - vars.zetah / (grid.Krsq .+ params.kdef^2)
```
navidcy, line 190: with the `@.` in front you don't need to do `.+` in the denominator

src/singlelayerqg.jl
```diff
@@ -218,7 +223,7 @@ addforcing!(N, sol, t, clock, vars::Vars, params, grid) = nothing
 function addforcing!(N, sol, t, clock, vars::ForcedVars, params, grid)
   params.calcF!(vars.Fh, sol, t, clock, vars, params, grid)
   @. N += vars.Fh
-  
+
```
navidcy, line 226: I think this is just Atom's `whitespace` package adding/removing spaces and make it as if you've modified more than you did... I had the same problem previously... @liasiegelman, perhaps you can disable `whitespace` package from Atom?

src/singlelayerqg.jl
```diff
@@ -245,7 +250,8 @@ Update the variables in `vars` with the solution in `sol`.
 """
 function updatevars!(sol, vars, params, grid)
   @. vars.zetah = sol
-  @. vars.psih  = -grid.invKrsq * vars.zetah
+  @. vars.psih  = - vars.zetah / (grid.Krsq .+ params.kdef^2)
```
navidcy, line 253: again, no `.+` in the denominator

src/singlelayerqg.jl
navidcy: Btw there is a problem here, if `kdef=0` then this will result to some NaNs!

src/singlelayerqg.jl
```diff
@@ -270,37 +276,63 @@ on the `grid`.
...
 set_zeta!(prob, zeta) = set_zeta!(prob.sol, prob.vars, prob.params, prob.grid, zeta)
 
 
 """
-    energy(prob)
-    energy(sol, grid)
+    k_energy(prob)
+    p_energy(prob)
+    k_energy(vars, grid)
+    p_energy(vars, grid, params)
```
navidcy, line 294: Let's split the docstrings. We should have one for kinetic energy and one for potential energy.

Also, may I suggest the rather more verbose `kinetic_energy` and `potential_energy` for names of these functions?

(Let us not be haunted by ghosts of the past from the old-fortran times when variable names had to be restricted to a maximum of 8 or 12 characters... :))

singlelayerqg.jl
```diff
@@ -245,7 +252,10 @@ Update the variables in `vars` with the solution in `sol`.
 """
 function updatevars!(sol, vars, params, grid)
   @. vars.zetah = sol
-  @. vars.psih  = -grid.invKrsq * vars.zetah
+  @. vars.psih  = - vars.zetah / (grid.Krsq + params.kdef^2)
+  if params.kdef == 0.0
```
navidcy, line 256: adding an `if` statement here might have ramifications in performance...

the more I think about it I lean towards defining a `streamfunctionfrompv!()` function and overload it, e.g.:

```julia
function streamfunctionfrompv!(vars, params::BarotropicParams, grid)
  @. vars.psih =  - grid.invKrsq * vars.sol
end

function streamfunctionfrompv!(vars, params::EquivalentBarotropicParams, grid)
  @. vars.psih  = - vars.sol / (grid.Krsq + params.kdef^2)
end
```

@glwagner, what do you reckon?

singlelayerqg.jl
```diff
@@ -1,15 +1,17 @@
-module BarotropicQG
+module SingleLayerQG
 
 export
   Problem,
   set_zeta!,
   updatevars!,
 
-  energy,
```
navidcy, line 8: Does it make sense to have `energy()` function giving the total energy of the flow? For a purely barotropic setup that would be simply the kinetic energy, when the deformation wavenumber is non-zero then energy would be the sum of kinetic + potential energy. We can do this by having some `AbstractTypes` for `Params`, e.g., something like

```Julia
energy(prob) = energy(prob.vars, prob.params, prob.grid)
energy(vars, params::BarotropicParams, grid) = kinetic_energy(vars, params, grid)
energy(vars, params::EquivalentBarotropicParams, grid) = kinetic_energy(vars, params, grid) + potential_energy(vars, params, grid)
```

navidcy: (All previously `BarotropicQG` examples fail to run because they are calling `SingleLayer.energy` function which does not exist... 5095b3d added a temp fix to that.)

navidcy:
@liasiegelman, food for thought: how about instead of `kdef` we use the more verbose `deformation_radius`. Then what you've been calling `kdef` will be simply `1 / deformation_radius`. We can have `deformation_radius = Inf` as the default value for a purely barotropic model.

singlelayerqg.jl
navidcy: 4650164 and 049bad0 resolve this.

singlelayerqg.jl
navidcy: resolved with 4650164 and 049bad0

src/singlelayerqg.jl
```diff
@@ -96,7 +96,7 @@ struct Params{T, Aphys, Atrans, ℓ} <: SingleLayerQGParams
 end
 
 const BarotropicQGParams = Params{<:AbstractFloat, <:AbstractArray, <:AbstractArray, Nothing}
-const EquivalentBarotropicQGParams = Params{<:AbstractArray, <:AbstractArray, <:AbstractArray, <:AbstractFloat}
+const EquivalentBarotropicQGParams = Params{<:AbstractFloat, <:AbstractArray, <:AbstractArray, <:AbstractFloat}
```
navidcy, line 99: nice catch 👍🏼

navidcy:
Initially `BarotropicQG` used `vars.ζ` to denote relative vorticity `∇²ψ`. With the extension to allowing for finite Rossby radius of deformation now `vars.ζ` is used to store the relative vorticity + vortex stretching term, i.e.,

```julia
vars.ζ = ∇²ψ - ψ/ℓ²
```

This could bring up some confusion...

In `MultilayerQG`, variable `vars.q` denotes the sum of relative vorticity + vortex stretching terms (https://fourierflows.github.io/GeophysicalFlowsDocumentation/dev/modules/multilayerqg/) while here `vars.q` was used to denote the sum of relative vorticity + topographic PV: `vars.q = ∇²ψ + η`

I suggest we change to use `vars.q` for
```julia
vars.q = ∇²ψ - ψ/ℓ²
```

Then we might even consider dropping `vars.ζ` whatsoever. Also, we don't need to have a separate variable for 
`∇²ψ - ψ/ℓ² + η` but rather write the sum `vars.q + params.eta` explicitly every time.

What do you think? @liasiegelman, @glwagner

src/singlelayerqg.jl
```diff
@@ -109,28 +132,34 @@ end
...
   CUDA.@allowscalar L[1, 1] = 0
   return FourierFlows.Equation(L, calcN!, grid)
 end
 
+function Equation(params::EquivalentBarotropicQGParams, grid::AbstractGrid)
+  L = @. - params.μ - params.ν * grid.Krsq^params.nν + im * params.β * grid.kr / (grid.Krsq + 1 / params.deformation_radius^2)
+  CUDA.@allowscalar L[1, 1] = 0
+  return FourierFlows.Equation(L, calcN!, grid)
+end
+
```
navidcy, line 146: @glwagner any suggestions for reducing the boilerplate in the Equation constructor?

liasiegelman:
@navidcy I agree that vars.ζ = ∇²ψ - ψ/ℓ² is a bit confusing. What you just suggested sounds good to me. Just one question, would defining  vars.q = ∇²ψ - ψ/ℓ² + η slow down the computation ?

navidcy:
@liasiegelman, defining `vars.q = ∇²ψ - ψ/ℓ² + η` is also fine. But if we'll be using `vars.q` to invert to get the stream function via `streamfunctionfrompv!()` then we'll need to subtract each time the topographic PV, right?
```julia
streamfunctionfrompv!(vars.ψh, vars.qh .- params.etah, params, grid)
```

Of course one could suggest that we just do
```julia
streamfunctionfrompv!(vars.ψh, sol, params, grid)
```
but that's a bit less verbose.
 
What I'm saying is that the code is now,

```julia
sol = ∇²ψ - ψ/ℓ²
```

so we should have one variable to correspond to what `sol` is. Previously that was ` vars.ζ` now should it be `vars.q`? Something else?

liasiegelman:
I think vars.q is fine

src/singlelayerqg.jl
```diff
@@ -0,0 +1,509 @@
...
+Returns the domain-averaged reduced enstrophy ½ ∫(q² + 2qη) dxdy / (Lx Ly) .
+```math
+\\int \\frac1{2} (q^2 + 2 q \eta) \\frac{\\mathrm{d}^2 \\boldsymbol{x}}{L_x L_y} = \\sum_{\\boldsymbol{k}} \\frac1{2} |\\hat{q}|^2 + 2 \hat{q} \hat{\eta} \\ .
+```
+"""
+function reduced_enstrophy(sol, vars, params, grid)
+  @. vars.qh = sol
+  return parsevalsum(abs2.(vars.qh) .+ 2 * vars.qh .* params.etah, grid) / (2 * grid.Lx * grid.Ly)
+end
```
navidcy, line 399: @liasiegelman,

1) what is this quantity "reduced enstrophy"? I've never encountered it elsewhere. (I'm just wondering if it's not something of general interest then perhaps we shouldn't include it in the module but the users can define it as a custom diagnostic?)

2) I'm pretty sure there is a `conj()` missing from the definition? I would expect that `∫ a b dx = ∑ conj(â)*b̂`...

navidcy: Both `enstrophy` and `reduced_enstrophy` only differ by a constant, correct?
```julia
reduced_enstrophy = enstrophy - ∫ η² d²x / (Lx * Ly)
```

liasiegelman: @navidcy Hi Navid, you are right. It's just a name we came up with Bill to look at the variable part of the enstrophy. It should probably not be implemented in the module.

navidcy: 354d613

navidcy:
@glwagner, me and @liasiegelman have worked on this PR. We feel it's time for somebody else to have a look. What do you think?

@liasiegelman we should add an example in the scripts with finite deformation radius; doesn't need to be anything too fancy. Do you have any idea? I thought perhaps we run simultaneously  a decaying 2D turbulence (β=0) with `deformation_radius=Inf` and `deformation_radius=2.0` and plot the relative vorticity of the two side-by-side.

navidcy:
@glwagner, there is no preview of docs (something related with Documenter ssh keys and what not). But you can preview the docs locally if you run
```bash
julia --project=docs/ -e 'using Pkg; Pkg.instantiate(); Pkg.develop(PackageSpec(path=pwd()))'; julia --project=docs/ docs/make.jl; open docs/build/index.html
```
from the repo's home directory.

navidcy:
singlelayerqg_barotropic_equivalentbarotropic

liasiegelman:
@navidcy that's neat. The title for the right panel should probably read \Lap \psi - psi/L_d^2 if it is sol that you are plotting

navidcy:
The animation indeed shows ∇²ψ for both cases.

liasiegelman:
oki :)

glwagner:
I'll take a look!

src/singlelayerqg.jl
```diff
@@ -0,0 +1,508 @@
...
+                   μ :: T            # Linear drag
+                   ν :: T            # Viscosity coefficient
+                  nν :: Int          # Hyperviscous order (nν=1 is plain old viscosity)
+              calcF! :: Function     # Function that calculates the forcing on QGPV q
+end
+
+const BarotropicQGParams = Params{<:AbstractFloat, <:AbstractArray, <:AbstractArray, Nothing}
+const EquivalentBarotropicQGParams = Params{<:AbstractFloat, <:AbstractArray, <:AbstractArray, <:AbstractFloat}
+
+get_topographicPV_grid_values(eta::Function, grid::AbstractGrid{T, A}) where {T, A} = A([eta(grid.x[i], grid.y[j]) for i=1:grid.nx, j=1:grid.ny])
```
glwagner, line 100: Maybe out of scope, but this works for any function of x, y (not just topographic PV), so you could put this in FourierFlows `utils.jl` or something and instead call it

```julia
on_grid(func, grid::TwoDimGrid{T, A}) where {T, A} = ...
```

src/singlelayerqg.jl
```diff
@@ -0,0 +1,508 @@
...
+"""
+    BarotropicQGParams(grid::TwoDGrid, β, eta, μ, ν, nν::Int, calcF
+
+Constructor for BarotropicQGParams (infinite Rossby radius of deformation).
+"""
+function BarotropicQGParams(grid::AbstractGrid{T, A}, β, eta, μ, ν, nν::Int, calcF) where {T, A}
+  eta_grid = typeof(eta) <: AbstractArray ? eta : get_topographicPV_grid_values(eta, grid)
+  eta_gridh = rfft(eta_grid)
+  return Params(β, nothing, eta_grid, eta_gridh, μ, ν, nν, calcF)
+end
```
glwagner, line 111: This could be

```julia
BarotropicQGParams(grid::AbstractGrid{T, A}, β, eta, μ, ν, nν::Int, calcF) where {T, A} =
    EquivalentBarotropicQGParams(grid, β, nothing, eta, μ, ν, nν, calcF)
```

src/singlelayerqg.jl
```diff
@@ -0,0 +1,508 @@
...
+  ldiv!(vars.v, grid.rfftplan, deepcopy(vars.vh))
+
+  return nothing
+end
+
+updatevars!(prob) = updatevars!(prob.sol, prob.vars, prob.params, prob.grid)
+
+"""
+    set_q!(prob, q)
+    set_q!(sol, vars, params, grid)
```
glwagner, line 302: I guess it makes sense just to document the user API here via `set_q!(prob, q)` rather than also including the "lower-level" form.

glwagner: Note there's also a typo in the lower level form, should be `set_q!(sol, vars, params, grid, q)`

glwagner (approved):
You've done a great job with this! Great work all!

src/singlelayerqg.jl
glwagner: Shenanignas probably can't save much code here so I think its ok.

src/singlelayerqg.jl
navidcy: You are suggesting that `set_q!(sol, vars, params, grid, q)` is useless and we just need to have `set_q!(prob, q)`? I think I agree... I don't remember where `set_q!(sol, vars, params, grid, q)` was useful for... perhaps legacy now...

src/singlelayerqg.jl
navidcy: Good suggestion. Let's do it in a separate PR.

PR #162, opened 2020-12-09
fixes previous issues with SingleLayerQG.jl

liasiegelman:


navidcy:
@liasiegelman let's just keep working on #160; no need to open a new PR.
I presume there will be more changes/suggestions to come anywayzz :)

PR #184, opened 2021-01-10
Minor enhancements in Docs/Stochastic Forcing section

navidcy:
- Fixes warning/complaining of Docs for duplicate ref to `TwoDNavierStokes.energy_work()`.
- Adds definition for Brownian motion `W_t`.
- Some code beautifications.

Screen Shot 2021-01-11 at 7 38 33 am

### DIFF
--- a/docs/src/stochastic_forcing.md
+++ b/docs/src/stochastic_forcing.md
@@ -521,7 +521,7 @@ stochastic forcing, remains the same. We can easily verify this from the "formal
 of (10):
 
 ```math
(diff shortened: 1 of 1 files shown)


glwagner:
Do we use subscript notation for derivatives anywhere? Could confuse the definition of W_t, which is not a time-derivative.

navidcy:
> Do we use subscript notation for derivatives anywhere? Could confuse the definition of W_t, which is not a time-derivative.

I don’t think we do use subscripts for derivatives...

liasiegelman:
I believe W_t is the standard notation for a Wiener process. You may want to add a fourth property though : 4. Continuity. With probability 1, W_t is continuous in t.

navidcy:
Note the sentence before properly 1.

liasiegelman:
oki :)

navidcy:
so shall I merge?
feel free to critique the **whole** Stochastic Forcing section if you like ;)

glwagner:
> so shall I merge?
> feel free to critique the **whole** Stochastic Forcing section if you like ;)

That was the only "issue" I found after a quick scan. Merge!

PR #189, opened 2021-01-26
Suggested updates to JOSS paper

BrodiePearson:
Here are some suggested updates to the JOSS paper Navid sent around. I just expanded or emphasized some points that I thought were important - feel free to remove or edit those additions. The JOSS website also said to 

> Mention (if applicable) a representative set of past or ongoing research projects using the software and recent scholarly publications enabled by it

So I mentioned a manuscript I currently have under review and some other work in progress. If there are other references demonstrating the package's utility you should add them!

[Here is the PDF](http://res.cloudinary.com/hju22ue2k/image/upload/v1611703165/agfkyufu9m6srokr5w2b.pdf)

### DIFF
--- a/paper/paper.md
+++ b/paper/paper.md
@@ -70,15 +70,15 @@ and refer to \autoref{eq:fourier} from text.
 
 # Summary
 
-`GeophysicalFlows.jl` is a Julia package that contains partial differential solvers for a collection of different large-scale geophysical fluid systems on periodic domains. All modules use Fourier-based pseudospectral numerical methods and leverage the 
+`GeophysicalFlows.jl` is a Julia package that contains partial differential solvers for a collection of large-scale geophysical fluid systems on periodic domains. All modules use Fourier-based pseudospectral numerical methods and leverage the 
 framework provided by the `FourierFlows.jl` Julia package for time-stepping, diagnostics, and 
 output.
 
 `GeophysicalFlows.jl` utilizes Julia's functionality and abstraction to enable all modules to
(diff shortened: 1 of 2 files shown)


paper/paper.md
```diff
@@ -70,20 +70,17 @@ and refer to \autoref{eq:fourier} from text.
 
 # Summary
 
-`GeophysicalFlows.jl` is Julia package that gathers together a collections of pseudospectral 
-solvers for geophysical fluid dynamics-related partial differential equations on periodic 
-domains. All modules use Fourier-based pseudospectral numerical methods and leverage the 
+`GeophysicalFlows.jl` is a Julia package that contains partial differential solvers for a collection of large-scale geophysical fluid systems on periodic domains. All modules use Fourier-based pseudospectral numerical methods and leverage the
```
navidcy, line 73: I'm not sure about "large-scale" in "large-scale geophysical fluid systems"...
Sure when we write these modules we are thinking of large-scale processes but, e.g., the `TwoDNavierStokes` module can be used to model fluid on a soap bubble... :)

navidcy (approved):

paper/paper.md
BrodiePearson: Very true, 2D dynamics/turbulence is a broad field. I wasn't sure how to link that in since the module's name is implies geophysics only. I used large-scale to distinguish from non-hydrostatic 3D ocean and atmosphere flows, but you should remove it if you like (I think I also said large-scale later)

navidcy: Yes, but later it was referring to QG modules so it's relevant there :)
OK, I'll remove this and merge. :)

PR #221, opened 2021-04-06
Fix some bugs

navidcy:
This PR fixes some bugs that came up with the way `CUDA.@allowscalar` is called: seems like if you are creating an array you need to call `CUDA.@allowscalar` *after* the `=` sign.


navidcy:
I'll merge as soon as tests pass since #219 relies on this.

PR #225, opened 2021-04-12
Optimize calcF! in examples with stochastic forcing

### LINKED ISSUE #223: Use CUDA.randn(nx, ny) instead of cu(randn(nx, ny))
This way of creating a stochastic forcing realization is not very efficient.

https://github.com/FourierFlows/GeophysicalFlows.jl/blob/aa1acb11ea89d8fb389084963836d9fb2ffe4989/examples/twodnavierstokes_stochasticforcing.jl#L65-L72

What we do here is create  `ξ` on CPU and then transfer to GPU if `dev=GPU()`. What we should do is use `CUDA.randn()` when on GPU, e.g.,

```julia
random_normal = dev==CPU() ? rand : CUDA.rand

 function calcF!(Fh, sol, t, clock, vars, params, grid) 
   ξ = exp.(2π * im * random_normal(eltype(grid), size(sol))) / sqrt(clock.dt) 
   @CUDA.allowscalar ξ[1, 1] = 0
    
   @. Fh = ξ * sqrt(forcing_spectrum)
    
   return nothing 
 end 
```

navidcy:
Closes #223.


glwagner:
If that works, that's awesome!

examples/barotropicqgql_betaforced.jl
```diff
@@ -66,13 +66,15 @@ seed!(1234) # reset of the random number generator for reproducibility
...
+# First we make sure that if `dev=GPU()`, then `CUDA.rand()` function is called for random
+# numbers uniformy distributed between 0 and 1.
+random_uniform = dev==CPU() ? rand : CUDA.rand
+
 function calcF!(Fh, sol, t, clock, vars, params, grid)
-  ξ = exp.(2π * im * rand(eltype(grid), size(sol))) / sqrt(clock.dt)
-  
-  Fh .= ArrayType(dev)(ξ) .* sqrt.(forcing_spectrum)
-  
+  Fh .= sqrt.(forcing_spectrum) .* exp(2π * im * random_uniform(eltype(grid))) ./ sqrt(clock.dt)
```
glwagner, line 74: Looks good :-D

Unfortunately you may need different functions for `Fh::Array` and `Fh::CuArray`, because you may need to use `CUDA.sqrt`?

navidcy: OK, there are some still some issues which I will fix.

But on the other hand, `CUDA.sqrt()` wasn't required previously... why it should be now?

glwagner: Oh, I guess you're right:

```julia
julia> using CUDA

julia> a = CuArray(rand(3))
3-element CuArray{Float64,1}:
 0.3974296213138291
 0.46740283369767677
 0.5039702920081996

julia> b = sqrt.(a)
3-element CuArray{Float64,1}:
 0.6304201942465272
 0.6836686578289785
 0.7099086504672272
```

PR #229, opened 2021-04-14
Simplify examples

### LINKED ISSUE #33: @reexport using FourierFlows in GeophysicalFlows.jl?
Why don't we call this
https://github.com/FourierFlows/GeophysicalFlows.jl/blob/c3c4f2d42b9f05f68c6cbfccd1043729eb511357/src/barotropicqg.jl#L16-L18
inside GeophysicalFlows.jl and instead we call it inside each module?

navidcy:
This PR simplifies the examples. Now common functions from FFTW and CUDA packages are re-exported by FourierFlows. Also, FourierFlows namespace is reexported by GeophysicalFlows. Thus, `using GeophysicalFlows` suffices for running simulations. The users still need to load, e.g.,`Plots.jl` if they want to plot the output.

(Inspired by @pdebuyl's remark in [JOSS paper review](https://github.com/openjournals/joss-reviews/issues/3053#issuecomment-818964253).)

Closes #33.

### DIFF
--- a/docs/Project.toml
+++ b/docs/Project.toml
@@ -3,11 +3,10 @@ CUDA = "052768ef-5323-5732-b1bb-66c8b64840ba"
 Documenter = "e30172f5-a6a5-5a46-863b-614d45cd2de4"
 FFTW = "7a1cc6ca-52ef-59f5-83cd-3a7055c09341"
 FourierFlows = "2aec4490-903f-5c70-9b11-9bed06a700e1"
-GeophysicalFlows = "44ee3b1c-bc02-53fa-8355-8e347616e15e"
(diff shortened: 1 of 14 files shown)


glwagner (approved):
Looks good.

Someday we may be able to combine functions like `energy` defined in different modules. We have to figure out a way to determine which physical problem a `FourierFlows.Problem` refers to (a simple way to achieve this would be to add type parameters to `FourierFlows.Problem` that identify the module it comes from...)

navidcy:
I guess if we do something like
```juliia
abstract type TwoDNavierStokesProblem <: AbstractProblem end
```
and define the `Problem` struct in FourierFlows.jl somehow to get the above subtype as a parameter?

PR #238, opened 2021-05-20
Fix `MultilayerQG` module on GPU & Julia v1.6

navidcy:


src/multilayerqg.jl
```diff
@@ -202,7 +202,7 @@ function Params(nlayers, g, f₀, β, ρ, H, U, eta, μ, ν, nν, grid; calcFq=n
     U = convert_U_to_U3D(dev, nlayers, grid, U)
 
   Uyy = real.(ifft(-l.^2 .* fft(U)))
-  Uyy = repeat(Uyy, outer=(nx, 1, 1))
+  Uyy = CUDA.@allowscalar repeat(Uyy, outer=(nx, 1, 1))
```
glwagner, line 205: Why `repeat`? Would `reshape` be enough?

glwagner (approved):

src/multilayerqg.jl
navidcy: `real.(ifft(-l.^2 .* fft(U)))` returns an `(1, ny, nlayers)` array and what we need is `(nx, ny, nlayers)`...

PR #250, opened 2021-06-04
Adds `aliased_fraction` kwarg in `Problem` constructors

navidcy:
This PR allows users to create module problems with different `aliased_fraction`.


navidcy:
@glwagner I'll merge because I want to fix the BCI example.

Issue #261, opened 2021-11-04
`Plots` no longer accepts functions as `clims`


glwagner:
When I run `examples/multilayerqg_2layer.jl` on a fresh clone on Julia 1.6.2 I get an error:

```julia
julia> include("multilayerqg_2layer.jl")
[ Info: Precompiling GeophysicalFlows [44ee3b1c-bc02-53fa-8355-8e347616e15e]
WARNING: could not import Printf.ini_hex into BFloat16s
WARNING: could not import Printf.ini_HEX into BFloat16s
[ Info: FourierFlows will use 8 threads
[ Info: FourierFlows will use 8 threads
[ Info: Precompiling Plots [91a5bcdd-55d7-5caf-9e0b-520d859cae80]
step: 0000, t: 0.0, cfl: 0.00, KE₁: 1.155e-08, KE₂: 5.483e-08, PE: 2.958e-09, walltime: 0.01 min
[ Info: Precompiling GR_jll [d2c73de3-f751-5644-a686-071e5b155ba9]
ERROR: LoadError: MethodError: no method matching iterate(::typeof(symlims))
Closest candidates are:
  iterate(::Union{LinRange, StepRangeLen}) at range.jl:664
  iterate(::Union{LinRange, StepRangeLen}, ::Int64) at range.jl:664
  iterate(::T) where T<:Union{Base.KeySet{var"#s79", var"#s78"} where {var"#s79", var"#s78"<:Dict}, Base.ValueIterator{var"#s77"} where var"#s77"<:Dict} at dict.jl:693
  ...
```

Is it the case that `Plots` no longer accepts function arguments to `clims`?

It's not so easy to fix the issue because some wizardry is used to update the plot data:

https://github.com/FourierFlows/GeophysicalFlows.jl/blob/dbd91a3a28ef9246c67e6f1f49e849640b8d3561/examples/multilayerqg_2layer.jl#L217-L220

discovered by @szy21

navidcy:
How about this for a fix?

```Julia
  for m in 1:nlayers
    clims = symlims(Array(vars.q[:, :, m]'))
    heatmap!(p[(m-1) * 3 + 1], x, y, Array(vars.q[:, :, m]'),
         aspectratio = 1,
              legend = false,
                   c = :balance,
               xlims = (-Lx/2, Lx/2),
               ylims = (-Ly/2, Ly/2),
               clims = clims,
              xticks = -3:3,
              yticks = -3:3,
[...]

PR #263, opened 2021-11-07
Pin CUDA.jl and Plots.jl

navidcy:
Quick fix to #261


navidcy:
https://github.com/JuliaPlots/Plots.jl/issues/3947

PR #270, opened 2021-11-20
Optimize 2-layer configs for `MultiLayerQG`

### LINKED ISSUE #267: Optimized PV inversion for two layer case in `MultilayerQG`
Currently we implement "optimized" functions for extracting PV from the QG streamfunction, and vice versa, in the single layer case:

https://github.com/FourierFlows/GeophysicalFlows.jl/blob/f4c68a58556f512c7118490fcb98679e68fff522/src/multilayerqg.jl#L485-L495

The inversion is also trivial in the two layer case, so we should support that.

To implement this I propose generalize `SingleLayerParams` to `NLayerParams{N, ...}`, where `N` is the number of layers. Then we have aliases

```julia
const SingleLayerParams = NLayerParams{1}
const TwoLayerParams = NLayerParams{2}
```

and we can implement optimized inversion routines for these two cases.

cc @szy21

navidcy:
This PR attempts to optimize the 2-layer configuration of `MultiLayerQG` module by hard-coding the PV-streamfunction relationships. To do so we create a `TwoLayerParams <: AbstractParams` and then use this type to add methods in both `streamfunctionfrompv!` and `pvfromstreamfunction!` functions that include the hard-coded PV-streamfunction relationships for 2-layer configurations.

Some benchmarks for 2- and 3-layer configs with `nx = ny = 128`:

with `nlayers=2`

**before this PR**
```julia
julia> using GeophysicalFlows, BenchmarkTools

julia> nlayers = 2; prob = MultiLayerQG.Problem(nlayers, GPU()); @btime stepforward!(prob)
3.317 s (700757 allocations: 114.42 MiB)
```

**after this PR**
```julia
julia> using GeophysicalFlows, BenchmarkTools

julia> nlayers = 2; prob = MultiLayerQG.Problem(nlayers, GPU()); @btime stepforward!(prob)
1.923 ms (3385 allocations: 321.58 KiB)
```

with `nlayers=3` (shouldn't be affected by this PR)

**before this PR**
```julia
julia> using GeophysicalFlows, BenchmarkTools

julia> nlayers = 3; prob = MultiLayerQG.Problem(nlayers, GPU()); @btime stepforward!(prob)
6.194 s (1299797 allocations: 213.44 MiB)
```

**after this PR**
```julia
julia> using GeophysicalFlows, BenchmarkTools

julia> nlayers = 3; prob = MultiLayerQG.Problem(nlayers, GPU()); @btime stepforward!(prob)
5.641 s (1299797 allocations: 213.44 MiB)
```

1650-fold faster for 2-layer config. We'll take that as an improvement. (Possibly even faster for higher grid resolution!)

Closes #267.

### DIFF
--- a/src/multilayerqg.jl
+++ b/src/multilayerqg.jl
@@ -228,6 +228,10 @@ struct TwoLayerParams{T, Aphys3D, Aphys2D, Trfft} <: AbstractParams
          ρ :: Aphys3D
     "array with rest height of each fluid layer"
          H :: Aphys3D
+    "rest height of top fluid layer"
+        H1 :: T
+    "rest height of bottom fluid layer"
+        H2 :: T
     "array with imposed constant zonal flow U(y) in each fluid layer"
          U :: Aphys3D
     "array containing topographic PV"
@@ -337,7 +341,7 @@ function Params(nlayers, g, f₀, β, ρ, H, U, eta, μ, ν, nν, grid; calcFq=n
     CUDA.@allowscalar @views Qy[:, :, nlayers] = @. Qy[:, :, nlayers] - Fm[nlayers-1] * (U[:, :, nlayers-1] - U[:, :, nlayers])
 
     if nlayers==2
-      return TwoLayerParams(T(g), T(f₀), T(β), A(ρ), A(H), U, eta, T(μ), T(ν), nν, calcFq, T(g′[1]), Qx, Qy, rfftplanlayered)
+      return TwoLayerParams(T(g), T(f₀), T(β), A(ρ), A(H), T(H[1]), T(H[2]), U, eta, T(μ), T(ν), nν, calcFq, T(g′[1]), Qx, Qy, rfftplanlayered)
     else # if nlayers>2
       return Params(nlayers, T(g), T(f₀), T(β), A(ρ), A(H), U, eta, T(μ), T(ν), nν, calcFq, A(g′), Qx, Qy, S, S⁻¹, rfftplanlayered)
     end
@@ -535,7 +539,7 @@ and `q2h = - k² * ψ2h + f₀² / (g′H₂) * (ψ1h - ψ2h)`.
 on the GPU.)
 """
 function pvfromstreamfunction!(qh, ψh, params::TwoLayerParams, grid)
-  f₀, g′, H1, H2 = params.f₀, params.g′, params.H[1], params.H[2]
+  f₀, g′, H1, H2 = params.f₀, params.g′, params.H1, params.H2
   
   ψ1h, ψ2h = ψh[:, :, 1], ψh[:, :, 2]
 
@@ -582,7 +586,7 @@ and `ψ2h = - k⁻² * q2h - (f₀² / g′) * (q1h / H₂ + q2h / H₁)`, where
 on the GPU.)
 """
 function streamfunctionfrompv!(ψh, qh, params::TwoLayerParams, grid)
-  f₀, g′, H1, H2 = params.f₀, params.g′, params.H[1], params.H[2]
+  f₀, g′, H1, H2 = params.f₀, params.g′, params.H1, params.H2
   
   q1h, q2h = qh[:, :, 1], qh[:, :, 2]
 
--- a/src/multilayerqg.jl
+++ b/src/multilayerqg.jl
@@ -107,10 +107,15 @@ function Problem(nlayers::Int,                        # number of fluid layers
         aliased_fraction = 1/3,
                        T = Float64)
 
-  if dev == GPU()
-    @warn """MultiLayerQG module not well optimized on the GPU yet.
-    See issue on Github at https://github.com/FourierFlows/GeophysicalFlows.jl/issues/112.
-    For now, we suggest running MultiLayerQG on CPUs only."""
+  if dev == GPU() && nlayers > 2
+    @warn """MultiLayerQG module not well optimized on the GPU yet for configurations with
+    3 fluid layers or more!
+    
+    See issues on Github at https://github.com/FourierFlows/GeophysicalFlows.jl/issues/112
+    and https://github.com/FourierFlows/GeophysicalFlows.jl/issues/267.
+    
+    To use MultiLayerQG with 3 fluid layers or more we suggest, for now, to restrict running
+    on CPU."""
   end
    
   # topographic PV
@@ -226,12 +231,8 @@ struct TwoLayerParams{T, Aphys3D, Aphys2D, Trfft} <: AbstractParams
          β :: T
     "array with density of each fluid layer"
          ρ :: Aphys3D
-    "array with rest height of each fluid layer"
-         H :: Aphys3D
-    "rest height of top fluid layer"
-        H1 :: T
-    "rest height of bottom fluid layer"
-        H2 :: T
+    "tuple with rest height of each fluid layer"
+         H :: Tuple
     "array with imposed constant zonal flow U(y) in each fluid layer"
          U :: Aphys3D
     "array containing topographic PV"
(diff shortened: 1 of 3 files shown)


navidcy:
90edadb is a temporary hack

navidcy:
Needs to be cleaned up. But seems to work!!

test/runtests.jl
```diff
@@ -23,6 +23,17 @@ const rtol_singlelayerqg = 1e-13 # tolerance for singlelayerqg forcing tests
...
+
+dev = GPU()
+
+@show nlayers = 2
+prob = GeophysicalFlows.MultiLayerQG.Problem(nlayers, dev)
+@btime stepforward!(prob)
+
+@show nlayers = 3
+prob = GeophysicalFlows.MultiLayerQG.Problem(nlayers, dev)
+@btime stepforward!(prob)
```
navidcy, line 36: I'll remove this before merging. I just put it here to make sure I don't introduce some slow-down shenanigans...

navidcy:
@glwagner @szy21 this is ready to be reviewed.

navidcy:
ok @glwagner and @szy21, feel free to review this!

src/multilayerqg.jl
```diff
@@ -211,6 +214,49 @@ struct SingleLayerParams{T, Aphys3D, Aphys2D, Trfft} <: BarotropicParams
...
+  # derived params
+    "the reduced gravity constants for the fluid interface"
+        g′ :: T
+    "array containing x-gradient of PV due to eta in each fluid layer"
+        Qx :: Aphys3D
+    "array containing y-gradient of PV due to β, U, and eta in each fluid layer"
+        Qy :: Aphys3D
+    "rfft plan for FFTs"
+  rfftplan :: Trfft
+end
```
glwagner, line 258: It seemed to me we could have `NLayerParams{N, ...}` to avoid copy-paste for `SingleLayerParams` (or even better, just add `N` to `Params`...) This is totally fine with me now if it works, we can clean up later.

src/multilayerqg.jl
```diff
@@ -455,42 +504,116 @@ Compute the inverse Fourier transform of `varh` and store it in `var`.
...
+(Here, the PV-streamfunction relationship is hard-coded to avoid scalar operations
+on the GPU.)
+"""
+function pvfromstreamfunction!(qh, ψh, params::TwoLayerParams, grid)
+  f₀, g′, H₁, H₂ = params.f₀, params.g′, params.H[1], params.H[2]
+  
+  ψ1h, ψ2h = ψh[:, :, 1], ψh[:, :, 2]
+
+  @. qh[:, :, 1] = - grid.Krsq * ψ1h + f₀^2 / (g′ * H₁) * (ψ2h - ψ1h)
+  @. qh[:, :, 2] = - grid.Krsq * ψ2h + f₀^2 / (g′ * H₂) * (ψ1h - ψ2h)
```
glwagner, line 555: Is there no need for `@views` here?

src/multilayerqg.jl
```diff
@@ -455,42 +504,116 @@ Compute the inverse Fourier transform of `varh` and store it in `var`.
...
+q̂₂ = - k² ψ̂₂ + f₀² / (g′ H₂) * (ψ̂₁ - ψ̂₂) .
+```
+
+(Here, the PV-streamfunction relationship is hard-coded to avoid scalar operations
+on the GPU.)
+"""
+function pvfromstreamfunction!(qh, ψh, params::TwoLayerParams, grid)
+  f₀, g′, H₁, H₂ = params.f₀, params.g′, params.H[1], params.H[2]
+  
+  ψ1h, ψ2h = ψh[:, :, 1], ψh[:, :, 2]
```
glwagner, line 552:
```suggestion
  ψ1h, ψ2h = view(ψh, :, :, 1), view(ψh, :, :, 2)
```

?

src/multilayerqg.jl
```diff
@@ -455,42 +504,116 @@ Compute the inverse Fourier transform of `varh` and store it in `var`.
...
+function streamfunctionfrompv!(ψh, qh, params::TwoLayerParams, grid)
+  f₀, g′, H₁, H₂ = params.f₀, params.g′, params.H[1], params.H[2]
+  
+  q1h, q2h = qh[:, :, 1], qh[:, :, 2]
+
+  @. ψh[:, :, 1] = - grid.Krsq * q1h - f₀^2 / g′ * (q1h / H₂ + q2h / H₁)
+  @. ψh[:, :, 2] = - grid.Krsq * q2h - f₀^2 / g′ * (q1h / H₂ + q2h / H₁)
   
+  for j in 1:2
+    @. ψh[:, :, j] *= grid.invKrsq / (grid.Krsq + f₀^2 / g′ * (H₁ + H₂) / (H₁ * H₂))
```
glwagner, line 614: Similar question re: `view` as above.

glwagner (approved):
Looks great to me! It seems possible to reuse one `Params` struct by adding a type parameter `N` corresponding to the number of layers. Then we can define the aliases

```julia
const SingleLayerParams = Params{1}
const TwoLayerParams = Params{2}
```

I don't see this as necessary for this PR though, just future cleanup.

navidcy:
> Looks great to me! It seems possible to reuse one `Params` struct by adding a type parameter `N` corresponding to the number of layers. Then we can define the aliases
> 
> 
> 
> ```julia
> 
> const SingleLayerParams = Params{1}
> 
> const TwoLayerParams = Params{2}
> 
> ```
> 
> 
> 
> I don't see this as necessary for this PR though, just future cleanup.

I agree! Can you open an issue with these ideas for a future cleanup?

src/multilayerqg.jl
navidcy: Not sure. Why should we?

src/multilayerqg.jl
```diff
@@ -297,12 +341,17 @@ function Params(nlayers, g, f₀, β, ρ, H, U, eta, μ, ν, nν, grid; calcFq=n
     end
     CUDA.@allowscalar @views Qy[:, :, nlayers] = @. Qy[:, :, nlayers] - Fm[nlayers-1] * (U[:, :, nlayers-1] - U[:, :, nlayers])
 
-    return Params(nlayers, T(g), T(f₀), T(β), A(ρ), A(H), U, eta, T(μ), T(ν), nν, calcFq, A(g′), Qx, Qy, S, S⁻¹, rfftplanlayered)
+    if nlayers==2
+      return TwoLayerParams(T(g), T(f₀), T(β), A(ρ), (T(H[1]), T(H[2])), U, eta, T(μ), T(ν), nν, calcFq, T(g′[1]), Qx, Qy, rfftplanlayered)
```
szy21, line 345: Just for my own understanding - Why would this make it much faster? It seems we are still calculating S and S-1 in lines 330-334 when nlayers==2, even though they are not needed?

szy21 (approved):
Thank you for doing this! It will make my life much easier:)

src/multilayerqg.jl
navidcy: @szy21, this part is not what makes the difference. Either way computing or not the S and S^{-1} matrices only happens once so that was not what was causing the slowdown.

But by defining a new AbstractParameter type for 2-layers then we can add a new method to `streamfunctionfrompv!` that works specifically for 2-layers.

See, e.g., this general one
https://github.com/FourierFlows/GeophysicalFlows.jl/blob/4c310b84072207d0b73d2091e40f01abfab8da98/src/multilayerqg.jl#L560-L572

versus this 2-layer specific one that has the inversion hard-coded in.
https://github.com/FourierFlows/GeophysicalFlows.jl/blob/4c310b84072207d0b73d2091e40f01abfab8da98/src/multilayerqg.jl#L586-L618

src/multilayerqg.jl
navidcy: Adding @views in both `streamfunctionfrompv!` and `pvfromstreamfunction!` reduced the allocations but only by just a bit. There are still some allocations (see, e.g., the comment at the top of the PR). Should we worry?

navidcy: e.g., 
```julia
julia> nlayers = 2; prob = MultiLayerQG.Problem(nlayers, GPU()); @btime stepforward!(prob)
2.026 ms (3713 allocations: 326.77 KiB)
```
after `@views` became
```julia
julia> nlayers = 2; prob = MultiLayerQG.Problem(nlayers, GPU()); @btime stepforward!(prob)
1.923 ms (3385 allocations: 321.58 KiB)
```

PR #276, opened 2021-11-23
Support SpecialFunctions 2.x

stevengj:
The only breaking change in SpecialFunctions 2.0 [was deleting](https://github.com/JuliaMath/SpecialFunctions.jl/pull/297) the `factorial(x::Real) = gamma(x+1)` function, and since you don't use `factorial` this shouldn't affect you.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -24,7 +24,7 @@ FFTW = "^1"
 FourierFlows = "^0.7.1"
 JLD2 = "^0.1, ^0.2, ^0.3, ^0.4"
 Reexport = "^0.2, ^1"
-SpecialFunctions = "^0.10, ^1"
+SpecialFunctions = "^0.10, ^1, 2"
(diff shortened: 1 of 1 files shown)


navidcy:
Thanks @stevengj. I tried to update yesterday but CUDA.jl had to downgrade to v2... I presume some of other deps have to updated the project compat entries.

Project.toml
```diff
@@ -24,7 +24,7 @@ FFTW = "^1"
 FourierFlows = "^0.7.1"
 JLD2 = "^0.1, ^0.2, ^0.3, ^0.4"
 Reexport = "^0.2, ^1"
-SpecialFunctions = "^0.10, ^1"
+SpecialFunctions = "^0.10, ^1, 2"
```
navidcy, line 27: should it be `, 2` or `, ^2`?

navidcy (approved):

Issue #20, opened 2019-05-31
Module that advects particles in two-dimensional flow


glwagner:
As requested by @natschil.

miniufo:
Just about to open a new issue on this, but found it here which is years ago.  I see there is a [PassiveTracerFlows.jl](https://github.com/FourierFlows/PassiveTracerFlows.jl).  Is it possible to make a similar **ParticleTrackingFlows.jl** for online Lagrangian particle tracking?

After a quick look at PassiveTracerFlows.jl, I find it is still in evolving the tracer in spectral space.  For tracking, we need to solve:
```
$$
d\mathbf(x) = \mathbf(v) dt
$$
```

which would be easier on grid points.  When calculate the nonlinear terms, I guess all the variables in spectral space will be transformed back to grid space.  This may make the tracking implementable.  [SpeedyWeather](https://speedyweather.github.io/SpeedyWeatherDocumentation/dev/particles/) is doing a similar job, but on a spherical geometry.

How about doing this offline, by saving the outputs at a relatively high frequency?  I am worrying about the stochastic forcing, which has no memory in time.  The offline tracking may not respect this fact and renders a longer memory in particle trajectories (by output frequency and interpolation between outputs).

PR #279, opened 2022-01-04
Add passive tracer advection-diffusion in 2D flows

BrodiePearson:
I added a passive tracer to the 2D module for some personal experiments, and I wondered if this would be a useful addition for others. This is related, but not identical, to Issue #20. I wasn't sure whether this module fit in `GeophysicalFlows.jl`, `PassiveTracersFlows.jl`, or another project (I want to expand to reacting tracers in the future).

I approached this problem using the multi-layer QG module's framework of a 3D solution (where the top layer is dynamics, and other layers are tracers). 

- A rough overview of the structural differences between the new tracer 2D module and the original no-tracer 2D module can be seen in commit `Adds tracer to new 2D dynamics module`. There were also numerous small modifications to the code after this commit.

The general properties of the module are:

- It solves advection-diffusion equations for tracers in a two-dimensional flow.
- It can advect an arbitrary number of tracers (`ntracers`). The examples I have in this PR use 3 tracers.
- There is a user-specified tracer diffusivity `κ` with order `nκ`.
- The tracers can be initialized alongside the initial vorticity field through a new `set_ζ_and_tracers!` function.
- The new module (`twodnavierstokeswithtracer.jl`) frames the solution (`sol`) as a `nk x nl x nlayers` array, where `nlayers = ntracers +1`. The first layer (`sol[:,:,1]`) is the vorticity and the lower layers are each of the tracers.
- The module is kind of slow (adding 3 tracers almost quadruples the simulation time). I think this is because of the `calcN!` step, and this could probably be made more efficient.
- It is not monotonic, it creates some small negative tracer concentrations from an initially positive/zero tracer field. I think this might be a common feature of spectral tracer advection schemes. Adding diffusion reduces the occurence of these negative values.
- I have added example setups (**see videos below**) for both decaying and stochastically-forced 2D turbulence, each with 3 tracers (2 initialized with Gaussians of varying shape, and a third initialized as a straight band of tracer). In both cases, the simulation dynamics are identical to the examples without tracers that already exist. 

## Decaying 2D turbulence (upper left two panels are identical to existing example)
https://user-images.githubusercontent.com/43418592/147993730-744c0d8d-3492-493b-9da2-2b41b4737494.mp4

## Stochastically-Forced 2D turbulence (upper left two panels are identical to existing example)
https://user-images.githubusercontent.com/43418592/147993529-c9748938-c254-4512-b504-d53f6b5d9f44.mp4

### Decaying 2D example without tracers
2D_Decaying_NoTracer

### Decaying 2D example with 3 tracers
2D_Decaying_Tracer

### Stochastically-forced 2D example without tracers
2D_Stochastic_NoTracer

### Stochastically-forced 2D example with 3 tracers
2D_Stochastic_Tracer


navidcy:
OMG! I totally forgot about this PR!

navidcy:
Really, really sorry!

navidcy:
This is very related with what's @jbisits been up to lately over in `PassiveTracerFlows.jl`, e.g., https://github.com/FourierFlows/PassiveTracerFlows.jl/pull/53

navidcy:
See https://fourierflows.github.io/PassiveTracerFlowsDocumentation/dev/literated/turbulent_advection-diffusion/

navidcy:
So @BrodiePearson this is great! The nice thing about this implementation compared to that done in https://github.com/FourierFlows/PassiveTracerFlows.jl/pull/53 is that here both tracers and flow are evolved simultaneously with the same time-stepping scheme and thus time-stepping can have the same accuracy.

I'm debating what's the best way forward here. There is a fair bit of overlap between modules `TwoDNavierStokes` and `TwoDNavierStokesTracer`. I think (but I'm not settled on this yet) that the best way forward is that this module belongs in PassiveTracerFlows.jl. Or perhaps we combine the two modules (`TwoDNavierStokes` and `TwoDNavierStokesTracer`)?

@glwagner would be keen to hear your opinion on this.

jbisits:
Sorry to jump into the thread here but I agree this is great!

BrodiePearson:
@navidcy No problem, I had forgotten about it too so thank you for reviving it (I was going to pick it up again over summer). 

I will be back at work next week and will take a look at @jbisits PR & think a bit more about your comments!

PR #294, opened 2022-07-06
Fix PE calculation bug in MultiLayerQG.energies

### LINKED ISSUE #292: Potential missing factor in `MultiLayerQG.energies`
I think there may be a missing `1 / sum(params.H)` factor in the potential energies part of `MultiLayerQG.energies`.

This is because in the n-layer case, multiplying the PV equations by $\psi_n H_n/H$ (with $H = \Sigma H_n$) and adding them together to form the energy equation gives potential energy terms that look like

$$\mathrm{PE_{j+1/2}} = \frac{1}{2} \frac{f_0^2}{g_{j+1/2}'H} \iint (\psi_j - \psi_{j+1})^2 \frac{dx dy}{Lx Ly}.$$

The total energy is nearly conserved as expected in two-layer decaying turbulence simulations if the `1 / sum(params.H)` is added.

Does this check? If so, I can open a pull request.

apaloczy:
This fixes #292 by adding a missing `1/sum(params.H)` factor to the potential energies in `MultiLayerQG.energies`.

### DIFF
--- a/src/multilayerqg.jl
+++ b/src/multilayerqg.jl
@@ -916,7 +916,7 @@ The kinetic energy at the ``j``-th fluid layer is
 while the potential energy that corresponds to the interface ``j+1/2`` (i.e., the interface 
 between the ``j``-th and ``(j+1)``-th fluid layer) is
 ```math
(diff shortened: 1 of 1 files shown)


navidcy:
This is great! I was thinking to add/modify a test to catch this. The tests at the moment have H=1 and thus were passing anyway?

apaloczy:
> This is great! I was thinking to add/modify a test to catch this. The tests at the moment have H=1 and thus were passing anyway?

Right, I think that's what was happening. How's this? I tried writing a test with `sum(params.H) == 10` to see if it works.

navidcy:
@apaloczy bump a patch release in Project.toml to v0.14.1

navidcy (approved):

PR #305, opened 2022-09-02
Update to latest grid refactor

navidcy:


jbisits:
I accidentally pushed to my fork.. I think I have `src/multilayerqg.jl`and `src/barotropicqgql.jl` done but have something on now so will have to get back to it later.

navidcy:
I think this should do the job. Have a look.

jbisits (approved):

PR #313, opened 2022-12-09
Option to specify topographic PV large-scale gradients in `MultiLayerQG`

### LINKED ISSUE #307: Option to specify PV gradients explicitly rather than computing them spectrally
In a two-layer `MultiLayerQG` configuration with topography based on the Phillips problem example from the docs, specifying the topographic PV gradient analytically instead of computing it spectrally avoids numerical problems at the boundaries for a planar bottom slope. This seems to be because of the discontinuities introduced by the FFTs.

So an option to specify PV gradients like `etax` and `etay` directly for simple idealized topographies could be helpful. A while ago I also experimented with specifying `Uyy` analytically for a Bickley jet in the upper layer. Maybe that's another desirable feature that could be implemented in the same way.

Any thoughts on how to go about this? Perhaps using multiple dispatch with the `Params` struct? I'd be happy to open a PR once we decide how to implement it.

### Example 1) Spectrally calculated $\eta_y = \mathrm{IFFT}[ il \times \mathrm{FFT}(h) ] \times f_0/H_2$ with a linear slope $h = sy$:

https://user-images.githubusercontent.com/4955404/192116919-2324ce79-67e9-4eda-9f16-0f799006daab.mp4

### Example 2) Directly prescribed $\eta_y = s \times f_0/H_2$:

https://user-images.githubusercontent.com/4955404/192117079-bd1c1ce2-536e-40e2-b712-5f223c2a395f.mp4

apaloczy:
This closes #307 by adding options to specify a nonperiodic component of the topographic PV gradients (for example a planar slope) in the `MultiLayerQG` module. Also added one test.

The gradients of the nonperiodic part of the topographic PV are now specified separately via `etax_nonperiodic` and `etay_nonperiodic` in `MultiLayerQG.Problem`, and are added to the gradients of the periodic part of the topographic PV (calculated via FFTs).

### DIFF
--- a/src/multilayerqg.jl
+++ b/src/multilayerqg.jl
@@ -72,9 +72,9 @@ Keyword arguments
   - `U`: The imposed constant zonal flow ``U(y)`` in each fluid layer.
   - `H`: Rest height of each fluid layer.
   - `ρ`: Density of each fluid layer.
-  - `eta`: Periodic component of the topographic potential vorticity.
-  - `etax_nonperiodic`: ``x``-gradient of the non-periodic component of the topographic potential vorticity.
-  - `etay_nonperiodic`: ``y``-gradient of the non-periodic component of the topographic potential vorticity.
+  - `eta`: Periodic component of the topographic potential vorticity. Default: `nothing`.
+  - `etax_nonperiodic`: ``x``-gradient of the non-periodic component of the topographic potential vorticity. Default: `nothing`.
+  - `etay_nonperiodic`: ``y``-gradient of the non-periodic component of the topographic potential vorticity. Default: `nothing`.
   - `μ`: Linear bottom drag coefficient.
   - `ν`: Small-scale (hyper)-viscosity coefficient.
   - `nν`: (Hyper)-viscosity order, `nν```≥ 1``.
@@ -85,6 +85,12 @@ Keyword arguments
   - `linear`: `true` or `false` (default); boolean denoting whether the linearized equations of motions are used.
   - `aliased_fraction`: the fraction of high-wavenumbers that are zero-ed out by `dealias!()`.
   - `T`: `Float32` or `Float64`; floating point type used for `problem` data.
+
+!!! note "Prescribing topographic potential vorticity"
+    Only the topographic potential vorticity gradients come into play into time-stepping forward the `MultiLayerQG`
+    system. The derivatives of the `eta` compoent are computed spectrally and, therefore, `eta` is expected to
+    be periodic. Any non-periodic topographic potential vorticity gradients can be prescribed via the `etax_nonperiodic`
+    and `etax_nonperiodic` keyword arguments.
 """
 function Problem(nlayers::Int,                        # number of fluid layers
                      dev = CPU();
--- a/src/multilayerqg.jl
+++ b/src/multilayerqg.jl
@@ -40,6 +40,8 @@ nothingfunction(args...) = nothing
                    H = 1/nlayers * ones(nlayers),
                    ρ = Array{Float64}(1:nlayers),
                  eta = nothing,
+    etax_nonperiodic = nothing,
+    etay_nonperiodic = nothing,
                    μ = 0,
                    ν = 0,
                   nν = 1,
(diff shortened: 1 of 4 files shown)


src/multilayerqg.jl
```diff
@@ -70,7 +72,9 @@ Keyword arguments
   - `U`: The imposed constant zonal flow ``U(y)`` in each fluid layer.
   - `H`: Rest height of each fluid layer.
   - `ρ`: Density of each fluid layer.
-  - `eta`: Topographic potential vorticity.
+  - `eta`: Periodic component of the topographic potential vorticity.
+  - `etax_nonperiodic`: x-gradient of the non-periodic component of the topographic potential vorticity.
+  - `etay_nonperiodic`: y-gradient of the non-periodic component of the topographic potential vorticity.
```
navidcy, line 77:
```suggestion
  - `etax_nonperiodic`: ``x``-gradient of the non-periodic component of the topographic potential vorticity.
  - `etay_nonperiodic`: ``y``-gradient of the non-periodic component of the topographic potential vorticity.
```

src/multilayerqg.jl
```diff
@@ -96,7 +100,9 @@ function Problem(nlayers::Int,                        # number of fluid layers
                        U = zeros(nlayers),            # imposed zonal flow U(y) in each layer
                        H = 1/nlayers * ones(nlayers), # rest fluid height of each layer
                        ρ = Array{Float64}(1:nlayers), # density of each layer
-                     eta = nothing,                   # topographic PV
+                     eta = nothing,                   # periodic component of the topographic PV
+        etax_nonperiodic = nothing,                   # x-gradient of the non-periodic component of the topographic PV
+        etay_nonperiodic = nothing,                   # y-gradient of the non-periodic component of the topographic PV
```
navidcy, line 105:
```suggestion
        etax_nonperiodic = nothing,                   # ``x``-gradient of the non-periodic component of the topographic PV
        etay_nonperiodic = nothing,                   # ``y``-gradient of the non-periodic component of the topographic PV
```

src/multilayerqg.jl
```diff
@@ -198,8 +214,12 @@ struct SingleLayerParams{T, Aphys3D, Aphys2D, Trfft} <: AbstractParams
          β :: T
     "array with imposed constant zonal flow ``U(y)``"
          U :: Aphys3D
-    "array containing topographic PV"
+     "array containing periodic component of the topographic PV"
```
navidcy, line 217:
```suggestion
     "array containing the periodic component of the topographic PV"
```

navidcy:
Sorry for slacking on this. I’m on holiday.

apaloczy:
No worries at all! Take your time.

navidcy (approved):

navidcy:
@apaloczy, I think it's good to go! Have a look at the docstrings additions I made and modify if needed.

navidcy:
I believe I gave you write rights so you can merge.

navidcy:
@apaloczy, I'm wondering whether there is a requirement that `etax_nonperiodic` and `etay_nonperiodic` are themselves periodic! I think they should! Otherwise, e.g., the `Qx` term won't be periodic and then errors will be introduced via

https://github.com/apaloczy/GeophysicalFlows.jl/blob/f51365a4587fd483bb7231418e26b3ec6d81bc66/src/multilayerqg.jl#L774-776

right?

navidcy (changes requested):
let's think more whether the gradients of the non-periodic component of the PV have to be periodic

navidcy:
Giving this more thought I doubt the non-periodic topo PV would work with anything but constant slope `η = a x + b y`.

glwagner:
The equations themselves have to be periodic; otherwise the discontinuity at domain "edges" will cause simulations to blow up.

navidcy:
I had a go at it. Have a look. :)

navidcy (approved):

apaloczy:
Sorry for the slowness here. I was offline for a couple of days myself. This all sounds great, I think having the large-scale topographic PV slopes as floats makes sense.

It also makes sense that `Qx` and `Qy` have to be periodic otherwise boundary errors will appear, and a planar slope is the only non-periodic topographic PV possible since it gives a periodic, constant PV gradient.

PR #326, opened 2023-06-09
Add derivation of discrete -> layered formulation in `Docs/MultiLayerQG`

navidcy:


docs/src/modules/multilayerqg.md
```diff
@@ -1,5 +1,85 @@
...
+```
+
+where ``N^2 = - \rho_0^{-1} g \, \mathrm{d} \rho / \mathrm{d}z`` is the square of the Brunt-Väisälä
+frequency for a Boussinesq fluid, where ``\rho_0`` is a reference density and ``g`` the gravitational
+constant.
+
+To go from the continuous (in ``z``) dynamics above to a discrete dynamics we can integrate all
+quantities over the height of each fluid layer.
+An alternative is to replace the vertical derivatives that appear both in the buoyancy definition and
+in the Brunt-Väisälä frequency with a discrete approximation.
```
glwagner, line 33: Can't we also make the QG approximation directly in Navier-Stokes written in buoyancy coordinates?

docs/src/modules/multilayerqg.md
```diff
@@ -1,5 +1,85 @@
...
+    &= (\partial_x^2 + \partial_y^2) \psi + \partial_z \left( \frac{f_0^2}{N^2} \partial_z \psi \right) .
+\end{aligned}
+```
+
+where ``N^2 = - \rho_0^{-1} g \, \mathrm{d} \rho / \mathrm{d}z`` is the square of the Brunt-Väisälä
+frequency for a Boussinesq fluid, where ``\rho_0`` is a reference density and ``g`` the gravitational
+constant.
+
+To go from the continuous (in ``z``) dynamics above to a discrete dynamics we can integrate all
+quantities over the height of each fluid layer.
```
glwagner, line 31: This is just a finite volume discretization of 3D QG, right?

docs/src/modules/multilayerqg.md
glwagner: I'm confused whether the "layered QG" refers to QG in buoyancy / isopycnal coordinates, or whether its merely the a more vanilla vertically-finite-volume discretization of 3D QG in z-coordinates.

docs/src/modules/multilayerqg.md
```diff
@@ -1,5 +1,85 @@
...
+
+We use subscripts ``j`` to denote quantities that are constant within each fluid layer. For example,
+the heights of each fluid layer are ``H_j``, ``j = 1, \dots, n``, where ``n`` is the number of fluid
+layers. Also, ``\rho_j`` is the density of each layer and ``\psi_j`` and ``q_j`` the streamfunction 
+and QGPV associated with that fluid layer.
+
+The buoyancy is associated with the fluid interfaces. We use semi-integer subscript values to denote a
+fluid interface, e.g., ``b_{5/2}`` is the buoyancy that corresponds to the interface between fluid layers
+2 and 3. Thus, from the relationship that connects buoyancy ``b`` and streamfunction ``\psi`` we approximate
+the buoyancies at each interface as
```
glwagner, line 43: This explanation suggests that we are really doing a transformation into buoyancy coordinates.

I'd suggest doing the transformation in continuous space first, ie write down QG in buoyancy coordinates. _Then_ you can introduce a "finite volume" approximation to QG in buoyancy coordinates. Since `b` is the coordinate, a finite volume discretization of `b` would indeed identify a buoyancy value for each layer.

navidcy:
Sorry. This meant to be a draft PR. I started it but then I wasn’t sure.. I’ll think about what you say.

PR #327, opened 2023-06-09
Fixing a sign error in Qy calc for multi-layer model

mjclobo:
There is a sign error on line 367 in multilayerqg.jl. The fix that I have made is consistent with the GeophysicalFlows.jl MultiLayerQG documentation equation. Fixing this bug greatly reduces the interior meridional PV gradient, params.Qy[1,1,2:nlayers-1], to a reasonable value for a prescribed background state.

There is still a second issue that keeps the interior PV gradient from being 0 when there is constant vertical shear, constant stratification, and no beta. I am talking to Navid now about possibly fixing this other issue.

I'm new to contributing on GitHub and am open to advice on best practices when creating pull requests. Thanks!


mjclobo:
Also, sorry for the 9 commits for 1 sign change fix. I was playing with both the sign change error and the definition of reduced gravity in order to verify that both changes are needed in order to have zero interior PV in an Eady case.

glwagner:
Can you show before/after plots that illustrate how this reduces the spurious interior PV? Along with code that generates the plots, if possible.

Issue #339, opened 2023-11-02
Mistake in stretching matrix in multilayerqg.jl


mpudig:
Hi,

We (Shafer Smith's group at NYU) noticed a possible typo in how the reduced gravities are defined in `multilayerqg` (https://github.com/FourierFlows/GeophysicalFlows.jl/blob/main/src/multilayerqg.jl#L355).

It is our understanding that in the multi-layer QG equations, one should divide by a reference density in the reduced gravities (see [Vallis, 2017](https://empslocal.ex.ac.uk/people/staff/gv219/aofd/) p.186), rather than by the density of each layer as is currently done in `multilayerqg` (as well as in its documentation).

@navidcy, could you comment on whether this is in fact a typo or if there was a reason behind this choice for the reduced gravities?

Thanks!

navidcy:
Hm... trying to remember.

Does the division with reference density implies that the Boussinesq approximation holds?

navidcy:
This [discussion](https://github.com/FourierFlows/GeophysicalFlows.jl/discussions/325) is probably relevant.

mpudig:
Sorry @navidcy, I hadn't seen this extensive discussion! We originally picked up this mistake in pyqg as Matt Lobo seems to have as well - ha! Happy to chat on that discussion. I believe that yes the reference density comes from Boussinesq.

navidcy:
No worries!

navidcy:
hi @mpudig, don't we still need to resolve whether we divide with $\rho_1$ or $\rho_n$?

mpudig:
Hi @navidcy, sorry for the slow uptake!

Is the conclusion from discussion #325 that we should do away with accepting $\rho$ as an input field for `multilayerqg` and only accept buoyancy? That way, if the stratification is constant then the reduced gravity will be constant layer-to-layer as we would expect.

`multilayerqg` will then not need any definition of $g$ or $\rho$ (nor a reference density) and will just require a buoyancy field input, $b$. I have had a tentative go at this change (see my PR).

If instead we do want to keep $\rho$ as an input, then yes I think we should simply divide by $\rho_1$ (the first layer value) to form the reduced gravity array (in order to adhere to the Boussinesq approximation and to avoid spurious internal PV gradients as Matt Lobo noted in the discussion).

Let me know what you think and happy to make changes as you see fit!

PR #343, opened 2023-11-20
Accepting buoyancy instead of density in `MultiLayerQG` module

mpudig:
This pull request addresses issue #339 and discussion #325.

The conclusion from the discussion was that we should accept buoyancy as an input instead of density in `multilayerqg`. The density array was used only to calculate the reduced gravity array. Previously, the derived reduced gravity array resulted in spurious internal PV gradients as discussed extensively in discussion #325. It also implied that the Boussinesq approximation did not hold, which is implicitly assumed in the form of the layered QG equations solved here.

This pull request has changed the input field of `multilayerqg` from $\rho$, the density, to $b = -g \frac{\delta \rho}{\rho_0}$, the Boussinesq buoyancy, where $|\delta \rho| \ll \rho_0$. The reduced gravity is now derived directly from $b$, https://github.com/mpudig/GeophysicalFlows.jl/blob/density_to_buoyancy/src/multilayerqg.jl#L348. I have set the change in buoyancy to scale like 1/nlayers with the other default parameters as they are in `multilayerqg` – I believe this is consistent with the other nondimensional parameter values and an order 1 Burger number.

### DIFF
--- a/src/multilayerqg.jl
+++ b/src/multilayerqg.jl
@@ -35,10 +35,9 @@ nothingfunction(args...) = nothing
                          Ly = Lx,
                          f₀ = 1.0,
                           β = 0.0,
-                          g = 1.0,
                           U = zeros(nlayers),
                           H = 1/nlayers * ones(nlayers),
-                          ρ = Array{Float64}(1:nlayers),
+                          b = -(1 .+ 1/nlayers * Array{Float64}(0:nlayers-1)),
                         eta = nothing,
     topographic_pv_gradient = (0, 0),
                           μ = 0,
@@ -67,10 +66,9 @@ Keyword arguments
   - `Ly`: Extent of the ``y``-domain.
   - `f₀`: Constant planetary vorticity.
   - `β`: Planetary vorticity ``y``-gradient.
-  - `g`: Gravitational acceleration constant.
   - `U`: The imposed constant zonal flow ``U(y)`` in each fluid layer.
   - `H`: Rest height of each fluid layer.
-  - `ρ`: Density of each fluid layer.
+  - `b`: Boussinesq buoyancy of each fluid layer.
   - `eta`: Periodic component of the topographic potential vorticity.
   - `topographic_pv_gradient`: The ``(x, y)`` components of the topographic PV large-scale gradient.
   - `μ`: Linear bottom drag coefficient.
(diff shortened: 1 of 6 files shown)


navidcy:
Thanks for this! I'll try to get to this sometime this week!

src/multilayerqg.jl
```diff
@@ -92,14 +90,13 @@ function Problem(nlayers::Int,                             # number of fluid lay
...
-                            U = zeros(nlayers),            # imposed zonal flow U(y) in each layer
-                            H = 1/nlayers * ones(nlayers), # rest fluid height of each layer
-                            ρ = Array{Float64}(1:nlayers), # density of each layer
-                          eta = nothing,                   # periodic component of the topographic PV
-      topographic_pv_gradient = (0, 0),                    # tuple with the ``(x, y)`` components of topographic PV large-scale gradient
+                           f₀ = 1.0,                                             # Coriolis parameter
+                            β = 0.0,                                             # y-gradient of Coriolis parameter
+                            U = zeros(nlayers),                                  # imposed zonal flow U(y) in each layer
+                            H = 1/nlayers * ones(nlayers),                       # rest fluid height of each layer
+                            b = -(1 .+ 1/nlayers * Array{Float64}(0:nlayers-1)), # Boussinesq buoyancy of each layer
```
navidcy, line 97: any reasoning behind this default `b`?

navidcy: is this what corresponds to the previous default ρ?

navidcy: I think it's the same if ρ0 = ρ[n], right?

navidcy (approved):
lgtm! nice job!

navidcy:
@mpudig you should have merge rights now, right?

if you are happy merge whenever.

navidcy:
@mpudig I'll merge

navidcy:
@mpudig welcome to GeophysicalFlows.jl contributors!

mpudig:
@navidcy Apologies for the delay, I was making the long trip back to Australia. Thanks for merging!!

src/multilayerqg.jl
mpudig: For the two-layer model this choice of default $b$ will give the same reduced gravity as the previous default $\rho$ did. I chose this as the default $b$ as I believe it's consistent with QG scalings.

PR #360, opened 2024-06-17
Add constant background flow into `SingleLayerQG`

mncrowe:
Added a constant background flow in the x-direction through a new parameter U of type T. This background advection is included in the linear term L of the equation.

### DIFF
--- a/src/singlelayerqg.jl
+++ b/src/singlelayerqg.jl
@@ -37,6 +37,7 @@ nothingfunction(args...) = nothing
                      Ly = Lx,
                       β = 0.0,
      deformation_radius = Inf,
+                      U = 0.0,
                     eta = nothing,
                       ν = 0.0,
                      nν = 1,
@@ -62,6 +63,7 @@ Keyword arguments
   - `Ly`: Extent of the ``y``-domain.
   - `β`: Planetary vorticity ``y``-gradient.
   - `deformation_radius`: Rossby radius of deformation; set `Inf` for purely barotropic.
+  - `U`: Background flow in the ``x``-direction.
   - `eta`: Topographic potential vorticity.
   - `ν`: Small-scale (hyper)-viscosity coefficient.
   - `nν`: (Hyper)-viscosity order, `nν```≥ 1``.
@@ -82,6 +84,7 @@ function Problem(dev::Device=CPU();
   # Physical parameters
                    β = 0.0,
   deformation_radius = Inf,
+                   U = 0.0,
                  eta = nothing,
   # Drag and (hyper-)viscosity
                    ν = 0.0,
@@ -103,7 +106,7 @@ function Problem(dev::Device=CPU();
   # topographic PV
   eta === nothing && (eta = zeros(dev, T, (nx, ny)))
 
-  params = deformation_radius == Inf ? BarotropicQGParams(grid, T(β), eta, T(μ), T(ν), nν, calcF) : EquivalentBarotropicQGParams(grid, T(β), T(deformation_radius), eta, T(μ), T(ν), nν, calcF)
+  params = deformation_radius == Inf ? BarotropicQGParams(grid, T(β), T(U), eta, T(μ), T(ν), nν, calcF) : EquivalentBarotropicQGParams(grid, T(β), T(deformation_radius), T(U), eta, T(μ), T(ν), nν, calcF)
 
   vars = calcF == nothingfunction ? DecayingVars(grid) : (stochastic ? StochasticForcedVars(grid) : ForcedVars(grid))
 
@@ -131,6 +134,8 @@ struct Params{T, Aphys, Atrans, ℓ} <: SingleLayerQGParams
                    β :: T
     "Rossby radius of deformation"
   deformation_radius :: ℓ
+    "Background flow in x direction"
+                   U :: T
     "topographic potential vorticity"
                  eta :: Aphys
     "Fourier transform of topographic potential vorticity"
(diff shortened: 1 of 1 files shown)


test/test_singlelayerqg.jl
```diff
@@ -435,4 +435,28 @@ function test_1layerqg_enstrophy_drag(dev; deformation_radius=2.23)
...
+	
+	prob = SingleLayerQG.Problem(dev; nx=n, Lx=L, dt, U = -1, stepper="FilteredRK4")
+	x, y = gridpoints(prob.grid)
+	
+	q₀ = lambdipole(1,1,prob.grid,center=(1e-10,0))
+	SingleLayerQG.set_q!(prob, q₀)
+	
+	stepforward!(prob, Int(5/dt))
+	
+	~,i₀ = findmax(q₀)
```
glwagner:
```suggestion
	_, i₀ = findmax(q₀)
```

`~` is an infix operator so we tend to use `_` as a neutral variable. Also mind the spelling.

test/test_singlelayerqg.jl
```diff
@@ -435,4 +435,28 @@ function test_1layerqg_enstrophy_drag(dev; deformation_radius=2.23)
...
+    test_1layerqg_background_flow(dev; n, L, dt)
+
+Evolves a lamb vortex with a background flow such that the vortex remains stationary.
+"""
+function test_1layerqg_background_flow(dev::Device=CPU(); n=256, L=10, dt=0.01)
+	
+	prob = SingleLayerQG.Problem(dev; nx=n, Lx=L, dt, U = -1, stepper="FilteredRK4")
+	x, y = gridpoints(prob.grid)
+	
+	q₀ = lambdipole(1,1,prob.grid,center=(1e-10,0))
```
glwagner, line 450:
```suggestion
	q₀ = lambdipole(1, 1, prob.grid, center=(1e-10, 0))
```

mind the spelling here

glwagner: :-D

https://www.sussex.ac.uk/informatics/punctuation/comma#:~:text=Each%20use%20has%20its%20own,followed%20by%20a%20white%20space.

test/test_singlelayerqg.jl
mncrowe: I think I already fixed this one...

mncrowe:
I've added a draft of the full functionality with `U = U(y)` (periodic in `y`). One question: when evaluating `CalcN!` there's an assignment `Uy = real.(ifft(im * grid.l .* fft(params.U)))`. This seems to be required as there's no 1D fft plan in the grid that could be used with `mul!`. My questions:

1. Is this the best way to do this derivative?
2. An identical value is calculated and assigned each timestep. As it's only in the `y` direction it shouldn't take too much memory, however, is this step (in particular, assigning GPU memory) likely to slow down the timestepping?
3. We could have `Uy` saved in `params` but this seems unnecessary. Do you agree?

src/singlelayerqg.jl
```diff
@@ -106,7 +106,13 @@ function Problem(dev::Device=CPU();
   # topographic PV
   eta === nothing && (eta = zeros(dev, T, (nx, ny)))
 
-  params = deformation_radius == Inf ? BarotropicQGParams(grid, T(β), T(U), eta, T(μ), T(ν), nν, calcF) : EquivalentBarotropicQGParams(grid, T(β), T(deformation_radius), T(U), eta, T(μ), T(ν), nν, calcF)
+  if U isa Number
+	U = T(U)
```
glwagner, line 110:
```suggestion
	U = convert(T, U)
```

just a little thing to help interpretability

src/singlelayerqg.jl
```diff
@@ -106,7 +106,13 @@ function Problem(dev::Device=CPU();
...
+  else
+	U = device_array(dev)(reshape(U,(1, grid.ny)))
```
glwagner, line 112:
```suggestion
	U = reshape(U, (1, grid.ny)))
	U = device_array(dev, U)
```

might be wrong but I think this is how it's supposed to be used?

also it doesn't seem necessary to require that `U` is an array (it could be a function too right? then it can change in time). But I think it's ok for this PR just to add this and make a comment that supporting additional types is future work

mncrowe:
```
help?> device_array
search: device_array

  device_array(device::Device)
  device_array(device::Device, T, dim)

  Return the proper array type according to the device, i.e., Array for CPU and CuArray for GPU.

  ────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────

  device_array(grid::AbstractGrid)

  Return the proper array type according to the grid's device, i.e., Array for CPU and CuArray for GPU.
```

navidcy:
@mncrowe is it OK if I push few edits?

navidcy (changes requested):
@mncrowe this looks very good! Thanks!

I changed the tests for when U is not varying with y to utilize the Rossby wave test.

There are two things left to do:

1.  remove the `if params.U isa Number ... else ...` statement from `calc_advection!`
 https://github.com/mncrowe/GeophysicalFlows.jl/blob/dd3c1ad4b6b3a3b2866525ad3ae993f3f3329270/src/singlelayerqg.jl#L352-L368

   which reduces performance; and

2. add a test for a background flow U(y). This can be done via the `test_1layerqg_nonlinearadvection` test, 

https://github.com/mncrowe/GeophysicalFlows.jl/blob/dd3c1ad4b6b3a3b2866525ad3ae993f3f3329270/test/test_singlelayerqg.jl#L257

   similarly to how this was done for the MultilayerQG module; see

https://github.com/mncrowe/GeophysicalFlows.jl/blob/dd3c1ad4b6b3a3b2866525ad3ae993f3f3329270/test/test_multilayerqg.jl#L215-L305

navidcy:
> There are two things left to do:
> 
> 1. remove the `if params.U isa Number ... else ...` statement from `calc_advection!`
>    https://github.com/mncrowe/GeophysicalFlows.jl/blob/dd3c1ad4b6b3a3b2866525ad3ae993f3f3329270/src/singlelayerqg.jl#L352-L368

ace06eb attempts to address 1.

I only wrote two methods for `calcN_advection!` that dispatch on the type of the param.Ul
I haven't checked whether the terms written there for non-zero U are correct. But doing a test as point 2 suggests will ensure that the terms are correct!

navidcy:
Btw, we also need to update `docs/src/modules/singlelayerqg.md`.

navidcy:
I think I see a minor problem. When `U` is just a number, don't we need a term `- U ∂η/∂x`? At the moment this term is not included.

I think I'll try to refactor the code to mimic the MultiLayerQG where the background PV gradients include the ∂η/∂x and ∂η/∂y terms and the ∂²U/∂y² (if U is not a constant).

navidcy:
@mncrowe I think I did everything! Have a look and let me know if you find any issues running the code or if the code is not behaving as you expect? Otherwise I'm happy to merge this!

navidcy:
I would just appreciate your look on the code. In principle I can approve the PR but it'd be nice if somebody else has a look because I made few changes.

navidcy (approved):

mncrowe:
Thanks for the very thorough reviewing and rewrite. Is there anything outstanding for me to do?

navidcy:
Does it work as expected?

navidcy:
A flow at rest advected by a background imposed U=0.5 over some topography.

```Julia
using GeophysicalFlows, GLMakie, Printf

    dev = CPU()
      n = 128
stepper = "FilteredRK4"
     dt = 0.04
 nsteps = 2000
 nsubs  = 5
      L = 2π


σx, σy = 0.4, 0.8
topographicPV(x, y) = 3exp(-(x - 1)^2 / 2σx^2 - (y - 1)^2 / 2σy^2) - 2exp(- (x + 1)^2 / 2σx^2 - (y + 1)^2 / 2σy^2)

prob = SingleLayerQG.Problem(dev; nx=n, Lx=L, eta=topographicPV, U=0.5, dt, stepper, aliased_fraction=0)
sol, clock, vars, params, grid = prob.sol, prob.clock, prob.vars, prob.params, prob.grid

x,  y  = grid.x,  grid.y
Lx, Ly = grid.Lx, grid.Ly

q = Observable(Array(vars.q))

fig = Figure(size=(380, 300))

axis_kwargs = (xlabel = "x", ylabel = "y", aspect = 1,
               limits = ((-Lx/2, Lx/2), (-Ly/2, Ly/2)))

title_q = Observable("initial vorticity ∂v/∂x-∂u/∂y")
axq = Axis(fig[1, 1]; title = title_q, axis_kwargs...)

hm = heatmap!(axq, x, y, q, colormap = :balance, colorrange = (-6, 6))

Colorbar(fig[1, 2], hm)

contour!(axq, x, y, params.eta;
         levels = collect(0.5:1:2.5), linewidth = 2, color = (:black, 0.5))
contour!(axq, x, y, params.eta;
         levels = collect(-2.5:1:-0.5), linewidth = 2, color = (:grey, 0.7), linestyle = :dash)

title_q[] = "vorticity, t=" * @sprintf("%.2f", clock.t)

startwalltime = time()

record(fig, "singlelayerqg_topography_U0.mp4", 0:round(Int, nsteps/nsubs), framerate = 20) do j
  q[] = irfft(sol, grid.nx)
  title_q[] = "vorticity, t=" * @sprintf("%.2f", clock.t)
  stepforward!(prob, nsubs)
end
```

https://github.com/user-attachments/assets/57838b1b-cbd3-4311-84da-6d2748ab1863

mncrowe:
> Does it work as expected?

All my scripts are producing the same results as before, and match my earlier results using Dedalus.

navidcy:
Nice! I’m merging then.

PR #366, opened 2024-09-07
Clarify `examples/README`

glwagner:
I thought it was confusing that the directory is `examples/` but the README says to navigate to `examples/directory`


navidcy (approved):
probably we should delete the .toml/readme.md from the examples directory all together and have clear instructions on how to run these in the docs

PR #371, opened 2024-10-21
Julia v1.11 for CI + update installation Julia version recommendations

navidcy:


jbisits:
I have had a bit of a search but have not been able to figure out what the appveyor error is

```terminal
  ✗ CUDA
  ✗ CUDA → SpecialFunctionsExt
  ✗ CUDA → ChainRulesCoreExt
  ✗ FourierFlows
        Info Given GeophysicalFlows was explicitly requested, output will be shown live 
ERROR: LoadError: ArgumentError: invalid Bool representation: "True"
```

Maybe due to package compats? Or something else that changed with  CUDA.jl?

navidcy:
Yeah, not sure.
Perhaps we should move away from appveyor and use [the GitHub actions](https://github.com/JuliaLang/Example.jl/blob/ee9d4f1ee0897a1258219631d43add681085f7f3/.github/workflows/ci.yml#L19) instead.

jbisits:
> Yeah, not sure. Perhaps we should move away from appveyor and use [the GitHub actions](https://github.com/JuliaLang/Example.jl/blob/ee9d4f1ee0897a1258219631d43add681085f7f3/.github/workflows/ci.yml#L19) instead.

I think that is good idea --- I use github actions on all of my (much smaller) projects.

navidcy:
@jbisits I think that should do the job ;)

jbisits:
Maybe also need to update the  [installation](https://github.com/FourierFlows/GeophysicalFlows.jl?tab=readme-ov-file#installation) section of README.md mentioning LTS is Julia v1.10 and this is minimum version code is tested on?

jbisits (approved):

Issue #112, opened 2020-09-15
Use KernelAbstractions to accelerate `MultilayerQG.streamfunctionfrompv!`


glwagner:
`KernelAbstractions.jl` can be used to accelerate the function 

https://github.com/FourierFlows/GeophysicalFlows.jl/blob/47a2b51def8af38b727dd64542c87051f169d5ea/src/multilayerqg.jl#L299-L302

A simple example showing how to use `KernelAbstractions` is the "Naive Transpose":

https://juliagpu.gitlab.io/KernelAbstractions.jl/examples/naive_transpose/

glwagner:
The first step is to write a kernel, which will look something like

```julia
@kernel invert_column!(ψh, qh, S⁻¹)
    i, j = @index(Global, NTuple)
    ψh .= S⁻¹ * qh
end
```

The next step is to create a work layout over which the kernel is launched. If we restrict attention to models that always have more than 32 grid points, we can use something like

```julia

# Larger workgroups are generally more efficient. For more generality, we could put an if statement that incurs
# different behavior when either nkl or nl are less than 16
workgroup = 16, 16

# The size determines how many times the kernel is run
worksize = grid.nkr, grid.nl

# This (and its useage below) will ensure the kernel is not run _before_ the data in qh is available
barrier = Event(dev)

# Creates a loop over the specified worksize, using workgroup to organize the computation
loop_invert_column! = invert_column!(dev, workgroup, worksize)

# Launch the kernel
event = loop_invert_column!(ψh, qh, params.invS, dependencies=barrier)

# This will ensure that no other operations occur until the kernel has finished
wait(dev, event)
```

glwagner:
One thing I am not totally sure about is whether `KernelAbstractions` will compile away the matrix multiplication in `ψh .= S⁻¹ * qh`. I _think_ that it will. If not, we may have to unroll our own loop.

glwagner:
By the way, I think this optimization also requires the columns of `ψh[i, j]` to be stored as `StaticArrays`. It looks like `ψh` is a 3D array right now. Other parts of the code may also have to converted to kernels if this change is made, since broadcasting over the 3D array would no longer work.

navidcy:
With this last suggestion would x, y FFTs work nicely?

glwagner:
> With this last suggestion would x, y FFTs work nicely?

Oof, good point.

Hmm, maybe we need to hand-write the matrix matrix multiply then. Not sure.

[...]

PR #373, opened 2024-10-28
Accelerating MultiLayerQG on GPUs + remove hardcoded 2-layer implementation

mpudig:
This pull request addresses accelerating the PV-stream function inversion in `MultiLayerQG` for arbitrary layers on a GPU, as discussed [here](https://github.com/FourierFlows/GeophysicalFlows.jl/issues/112).

I've used `KernelAbstractions` to optimize what used to be a loop over all _(x, y)_. There is still a loop over number of layers squared. These changes greatly accelerate the code for certain set-ups with more than two layers based on some simple tests (seen [here](https://github.com/FourierFlows/GeophysicalFlows.jl/issues/112#issuecomment-2256441301)).

Also the PR removes the hardcoded 2-layers implementation.


navidcy:
We should bump a patch release.

src/multilayerqg.jl
```diff
@@ -111,16 +113,16 @@ function Problem(nlayers::Int,                             # number of fluid lay
              aliased_fraction = 1/3,
                             T = Float64)
 
-  if dev == GPU() && nlayers > 2
-    @warn """MultiLayerQG module is not optimized on the GPU yet for configurations with
-    3 fluid layers or more!
+  # if dev == GPU() && nlayers > 2
+  #   @warn """MultiLayerQG module is not optimized on the GPU yet for configurations with
+  #   3 fluid layers or more!
```
navidcy, line 118: Why comment these out? Delete?

src/multilayerqg.jl
```diff
@@ -533,16 +535,50 @@ Compute the inverse Fourier transform of `varh` and store it in `var`.
...
+
+  # Instantiates the kernel for relevant backend device
+  backend = KernelAbstractions.get_backend(qh)
+  kernel! = pvfromstreamfunction_kernel!(backend, workgroup, worksize)
+
+  # Launch the kernel
+  S, nlayers = params.S, params.nlayers
+  kernel!(qh, ψh, S, nlayers)
+
+  # This will ensure that no other operations occur until the kernel has finished
```
navidcy, line 580:
```suggestion
  # ensure that no other operations occur until the kernel has finished
```

src/multilayerqg.jl
```diff
@@ -587,16 +623,50 @@ function pvfromstreamfunction!(qh, ψh, params::TwoLayerParams, grid)
...
+
+        @inbounds ψh[i, j, k] = 0
+
+        @unroll for m = 1:nlayers
+            @inbounds ψh[i, j, k] += S⁻¹[i, j][k, m] * qh[i, j, m]
+        end
+    
+    end
+end
+
```
navidcy, line 645: are the two kernel functions identical except the order they want their arguments?

mpudig: Looks like it yes. It'd probably make sense to write one kernel in the more general form.

navidcy: I think so! Would make the code more robust.

src/multilayerqg.jl
```diff
@@ -533,16 +535,50 @@ Compute the inverse Fourier transform of `varh` and store it in `var`.
...
+transform of the PV from the streamfunction `ψh` in each layer using `qh = params.S * ψh`.
+"""
+@kernel function pvfromstreamfunction_kernel!(qh, ψh, S, nlayers)
+  i, j = @index(Global, NTuple)
+
+  @unroll for k = 1:nlayers
+
+      @inbounds qh[i, j, k] = 0
+
+      @unroll for m = 1:nlayers
```
glwagner, line 551: The `@unroll` don't do anything unless `nlayers` is known at compile time (this requires using `Val`, but I don't know if it will speed anything up... it might).

src/multilayerqg.jl
```diff
@@ -533,16 +535,50 @@ Compute the inverse Fourier transform of `varh` and store it in `var`.
 """
 invtransform!(var, varh, params::AbstractParams) = ldiv!(var, params.rfftplan, varh)
 
+"""
+    @kernel function pvfromstreamfunction_kernel!(qh, ψh, S, nlayers)
+
+Kernel for GPU acceleration of PV from streamfunction calculation, i.e., obtaining the Fourier
+transform of the PV from the streamfunction `ψh` in each layer using `qh = params.S * ψh`.
+"""
+@kernel function pvfromstreamfunction_kernel!(qh, ψh, S, nlayers)
```
glwagner, line 544:
```suggestion
@kernel function pvfromstreamfunction_kernel!(qh, ψh, S, ::Val{nlayers}) where nlayers
```

for `@unroll` you have to do this, and also pass `Val(nlayers)` rather than `nlayers` into the kernel when launching it. I don't know if it will speed things up though. It might.

src/multilayerqg.jl
```diff
@@ -533,16 +524,50 @@ Compute the inverse Fourier transform of `varh` and store it in `var`.
...
+  @unroll for k = 1:nlayers
+
+      @inbounds a[i, j, k] = 0
+
+      @unroll for m = 1:nlayers
+          @inbounds a[i, j, k] += M[i, j][k, m] * b[i, j, m]
+      end
+  
+  end
+end
```
mpudig, line 545: I rewrote the kernel in more general form and added `Val`. The code has sped up slightly, but the 16-thread CPU still outperforms the GPU. Compare these benchmarks to what I showed [here](https://github.com/FourierFlows/GeophysicalFlows.jl/issues/112#issuecomment-2256441301)

- **GPU**
`nlayers = 12; nx = 512; prob = MultiLayerQG.Problem(nlayers, GPU(); nx); @btime stepforward!(prob)
  668.165 ms (2533 allocations: 191.19 KiB)`

- **CPU with 16 threads**
`nlayers = 12; nx = 512; prob = MultiLayerQG.Problem(nlayers, CPU(); nx); @btime stepforward!(prob)
  444.419 ms (113 allocations: 5.61 KiB)`

navidcy: Are you sure you are timing the [GPU properly](https://cuda.juliagpu.org/stable/development/profiling/)?

mpudig: `julia> nlayers = 12; nx = 512; prob = MultiLayerQG.Problem(nlayers, GPU(); nx); @benchmark CUDA.@sync CUDA.@time stepforward!(prob)`
  ```
0.681338 seconds (57.95 k CPU allocations: 2.419 MiB) (18 GPU allocations: 345.250 MiB, 0.04% memmgmt time)
  0.678481 seconds (2.58 k CPU allocations: 194.141 KiB) (16 GPU allocations: 297.156 MiB, 0.03% memmgmt time)
  0.676472 seconds (2.58 k CPU allocations: 194.141 KiB) (16 GPU allocations: 297.156 MiB, 0.03% memmgmt time)
  0.694825 seconds (2.58 k CPU allocations: 194.141 KiB) (16 GPU allocations: 297.156 MiB, 0.03% memmgmt time)
  0.678072 seconds (2.58 k CPU allocations: 194.141 KiB) (16 GPU allocations: 297.156 MiB, 0.03% memmgmt time)
  0.677693 seconds (2.58 k CPU allocations: 194.141 KiB) (16 GPU allocations: 297.156 MiB, 0.03% memmgmt time)
  0.678237 seconds (2.58 k CPU allocations: 194.141 KiB) (16 GPU allocations: 297.156 MiB, 0.04% memmgmt time)
  0.677198 seconds (2.58 k CPU allocations: 194.141 KiB) (16 GPU allocations: 297.156 MiB, 0.03% memmgmt time)
  0.676980 seconds (2.58 k CPU allocations: 194.141 KiB) (16 GPU allocations: 297.156 MiB, 0.03% memmgmt time)
  0.676189 seconds (2.58 k CPU allocations: 194.141 KiB) (16 GPU allocations: 297.156 MiB, 0.03% memmgmt time)
  0.678326 seconds (2.58 k CPU allocations: 194.141 KiB) (16 GPU allocations: 297.156 MiB, 0.03% memmgmt time)
  0.677010 seconds (2.58 k CPU allocations: 194.141 KiB) (16 GPU allocations: 297.156 MiB, 0.03% memmgmt time)
  0.677142 seconds (2.58 k CPU allocations: 194.141 KiB) (16 GPU allocations: 297.156 MiB, 0.04% memmgmt time)
  0.676321 seconds (2.58 k CPU allocations: 194.141 KiB) (16 GPU allocations: 297.156 MiB, 0.04% memmgmt time)
  0.678115 seconds (2.58 k CPU allocations: 194.141 KiB) (16 GPU allocations: 297.156 MiB, 0.03% memmgmt time)
  0.677461 seconds (2.58 k CPU allocations: 194.141 KiB) (16 GPU allocations: 297.156 MiB, 0.03% memmgmt time)
  0.678255 seconds (2.58 k CPU allocations: 194.141 KiB) (16 GPU allocations: 297.156 MiB, 0.03% memmgmt time)
  0.677168 seconds (2.58 k CPU allocations: 194.141 KiB) (16 GPU allocations: 297.156 MiB, 0.03% memmgmt time)
  0.677529 seconds (2.58 k CPU allocations: 194.141 KiB) (16 GPU allocations: 297.156 MiB, 0.03% memmgmt time)
BenchmarkTools.Trial: 8 samples with 1 evaluation.
 Range (min … max):  676.718 ms … 678.645 ms  ┊ GC (min … max): 0.00% … 0.00%
 Time  (median):     677.681 ms               ┊ GC (median):    0.00%
 Time  (mean ± σ):   677.765 ms ± 618.941 μs  ┊ GC (mean ± σ):  0.00% ± 0.00%

  █                     █  ██       █   █                 █   █  
  █▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁█▁▁██▁▁▁▁▁▁▁█▁▁▁█▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁█▁▁▁█ ▁
  677 ms           Histogram: frequency by time          679 ms <

 Memory estimate: 199.41 KiB, allocs estimate: 2660.
```

Seems to be roughly the same as above? Unless I'm misunderstanding what this benchmark is doing...

**Comment by navidcy** [comment]
@mpudig any idea why CI breaks?

File: src/multilayerqg.jl (hunk 6 above)
Comment (glwagner) [reply src/multilayerqg.jl:545]: That's disappointing...

The first thing I would try to figure out is whether this function is indeed the bottleneck. It might be better, infact, to simply benchmark this function in isolation.

I don't know if it matters but I saw the workgroup is 8, 8. Usually we use 16, 16.

I would also check 3 layers first perhaps. The double inner loop gets slower with more layers, and perhaps the computational costs scale differently on CPU vs GPU. That might give a clue.

The loop is over `k, m` --- which are the slowest (last) indices in `a` and `b`. That _could_ be an issue. If you can benchmark this operation in isolation, then you can experiment with new arrays where `k, m` are the fastest / first indices in `a` and `b`. This experiment would tell you what kind of slow down that's incurring.

Maybe the `@unroll` is not working for some reason. When I've seen stuff like this before, people have used matrix / linear algebra via `StaticArrays`, rather than explicit loops as used here. If you are just testing the kernel in isolation, transforming `a`, `b` to arrays of `StaticVector` could also be something to experiment with.

In general with performance engineering, one has to really be persistent and creative and test test test. To make this easier you want to extract out the function you're trying to optimize and work with a very idealized test case that also allows you to change the data structure (rather than working with a FourierFlows script). Think of this as research. If you find that you need to rearrange memory differently, then we can come back to GeophysicalFlows and see whether that is feasible or not.

Comment (mpudig) [reply src/multilayerqg.jl:545]: @glwagner @navidcy Apologies for the slow uptake! 

As you suggested I tested the `streamfunctionfrompv` function in isolation as this was indeed the bottleneck. I tested the suggestions you proposed and writing `a`, `b` in the kernel as 2d arrays of `StaticVector` led, by far, to the best performance improvement.

When `a`, `b` are defined as such, and `M` (which will be `S` or `S⁻¹` depending on the inversion direction) is a 2d array of `StaticArray,` the kernel is very elegant:

```julia
@kernel function PVinversion_kernel!(a, b, M, ::Val{nlayers}) where nlayers
    i, j = @index(Global, NTuple)

    @inbounds a[i, j] = M[i, j] * b[i, j]

end
``` 

I benchmarked this new kernel and structure of `a`, `b` and compared it to what this PR previously proposed (what I call "current" in the attached figure). The results are pretty impressive – 3 orders of magnitude speed-up in some cases! (These tests were done on a v100 GPU.) With the previous code the complexity scaled as `nlayers^4` (matrix-vector multiply + the double loop) whereas without the double loop it scales as `nlayers^2`-ish. The proposed change is also faster for the `nlayers = 2` case, which currently hardcodes the inversion step.  

streamfunctionfrompv_GPU_benchmarking

I'm not an expert on writing code for GPUs – how feasible do you think it would be to write all the variables required to step the model forward (`qh`, `ψh`, `q`, `ψ`, etc.) as arrays of `StaticVector`? It seems like it would require changing the fftw structure.

Let me know what you think about this. I'd be happy to try reorganise the code and implement it if you think it's a good idea. This sort of acceleration would be a huge boon for my research – and obviously would benefit others who want to run these sort of layered QG simulations as well.

mpudig:
> @mpudig any idea why CI breaks?

Hm, it's the two 3layer tests which fail, which isn't ideal... It's also only for Julia 1.11 that they fail – maybe a package inconsistency? I can't seem to parse where in the test it fails.

src/multilayerqg.jl
glwagner: > I'm not an expert on writing code for GPUs – how feasible do you think it would be to write all the variables required to step the model forward

Not true. You are an expert. That is why this PR was able to succeed.

navidcy: exactly!!

glwagner: > how feasible do you think it would be to write all the variables required to step the model forward (qh, ψh, q, ψ, etc.) as arrays of StaticVector?

I don't exactly understand what the intent is, can you spell out your design a little more clearly?

I believe that FFTs act on CuArray of Floats. If I understand you correctly, I believe that you would like to use `StaticVector` for vertical directions, is that right? However, this would mean that we do not have CuArray of Floats, we have CuArray of StaticVector. So I don't think this is compatible with FFTs.

I think the simplest solution is to keep the CuArray representation of all the array, but either to 1) allocate new CuArray of StaticVector, which are temporary variables for performing the inversion or 2) build the StaticVector _inside_ the inversion kernel. 

Number 2 seems actually easier to try right now. It is something like

```julia
@kernel function PVinversion_kernel!(a, b, M, ::Val{N}) where N
    i, j = @index(Global, NTuple)

    b_tuple = ntuple(Val(N)) do n
        @inbounds b[i, j, n]
    end
  
    T = eltype(a) 
    b_sv = SVector{N, T}(b_tuple)
    a_sv = @inbounds M[i, j] * b_sv

    ntuple(Val(N)) do n
        @inbounds a[i, j, n] = a_sv[n]
    end
end
```

There may be many creative ways to do this though. Possibly, it is possible to reinterpret the memory occupied by a CuArray of SVector as a simple CuArray with a different shape. There is a function `reinterpret` for this. Then perhaps you can perform ffts.

mpudig: > I don't exactly understand what the intent is, can you spell out your design a little more clearly?

@glwagner
I guess I assumed that if these fields needed to be `CuArray` of `StaticVector` at every time step then it would be more efficient for this to be their default memory structure – but I clearly didn't think about this very hard as your suggestion number 2 seems to do the trick with minimal rewriting! I tried a few riffs on this code but what you've written seems to do the job perfectly:

I timed `stepforward!` using this new code ("proposed") vs. what this PR initially suggested ("current"), and it scales as expected from what I showed above for `streamfunctionfrompv!`. Seems to be a success! Let me know if you have other thoughts.

stepforward_GPU_benchmarking

I'll have a dig around why the CI from the previous version of this PR was failing.

mpudig:
> > @mpudig any idea why CI breaks?
> 
> Hm, it's the two 3layer tests which fail, which isn't ideal... It's also only for Julia 1.11 that they fail – maybe a package inconsistency? I can't seem to parse where in the test it fails.

This newest commit to the PR seems to pass the tests that previously failed. Curious!

navidcy:
We should bump a patch release

src/multilayerqg.jl
glwagner: Amazing. Great job. Also Julia is great. Is there memory allocation incurred by the kernel or does it get compiled away in some magical way?

mpudig: This would be seen in the memory allocations posted by `@btime` right? This was what I found when timing `streamfunctionfrompv!`:

**Not using `SVector` in kernel**
- nlayers = 3, 6, 12, 24, 36, 48, 72: allocations = 51, 51, 97, 97, 99, 103, 134

**Using `SVector` in kernel**
- nlayers = 3, 6, 12, 24, 36, 48, 72: allocations = 51, 51, 51, 51, 97, 97, 97

glwagner: ok so the same? I think you are reporting the _number_ of allocations but not the total memory allocated (I think we are more concerned about the latter, even a single allocation can be destructive if it is huge)

mpudig: For the same calculation, total memory allocation looks pretty similar...

**Not using `SVector` in kernel**
- nlayers = 3, 6, 12, 24, 36, 48, 72: memory allocation (KiB) = 2.30, 2.30, 5.25, 5.25, 5.30, 5.39, 6.39

**Using `SVector` in kernel**
- nlayers = 3, 6, 12, 24, 36, 48, 72: memory allocation (KiB) = 2.19, 2.19, 2.19, 2.19, 5.14, 5.14, 5.14

Sorry if I'm misunderstanding and this isn't what you're asking!

navidcy:
@mpudig this is great!
should we remove the 2-layer Params and 2-layer inversion?
does the kernel inversion matches or beats that or you think it still worths keeping that?

src/multilayerqg.jl
mpudig: When I benchmark using

```julia
@benchmark CUDA.@sync CUDA.@time MultiLayerQG.streamfunctionfrompv!(ψh, qh, params, grid)
``` 
 as Navid suggested above, there is always a big (potentially very big) memory allocation initially but then every other sample is significantly smaller: e.g., for `nlayers = 72` and `nx = 128`:
```julia
@benchmark CUDA.@sync CUDA.@time MultiLayerQG.streamfunctionfrompv!(ψh, qh, params, grid)
 40.846772 seconds (20.30 M CPU allocations: 1.033 GiB, 1.19% gc time)
  0.008261 seconds (97 CPU allocations: 5.250 KiB)
  0.008244 seconds (98 CPU allocations: 5.547 KiB)
... etc
``` 
Is this what you mean by a single allocation being potentially destructive? And that the kernel memory allocation is magically compiled away?

mpudig:
@navidcy Looks good!

Re: keeping TwoLayerParams, timestepping may be _slightly_ faster when using the kernel inversion. See below some stats (all done on an rtx8000 with `nx = 128`):

**Two-layer using kernel inversion** 
- `@btime MultiLayerQG.streamfunctionfrompv!(ψh, qh, params, grid): 15.789 μs (51 allocations: 2.30 KiB)`
- `@btime MultiLayerQG.pvfromstreamfunction!(qh, ψh, params, grid): 15.590 μs (51 allocations: 2.30 KiB)`
- `@btime stepforward!(prob): 1.213 ms (2345 allocations: 179.31 KiB)`
- `@btime stepforward!(prob, 100): 124.988 ms (234500 allocations: 17.51 MiB)`

**Two-layer using hard-coded inversion** 
- `@btime MultiLayerQG.streamfunctionfrompv!(ψh, qh, params, grid): 77.866 μs (424 allocations: 42.25 KiB)`
- `@btime MultiLayerQG.pvfromstreamfunction!(qh, ψh, params, grid): 23.741 μs (158 allocations: 16.19 KiB)`
- `@btime stepforward!(prob): 1.203 ms (3837 allocations: 339.06 KiB)`
- `@btime stepforward!(prob, 100): 149.596 ms (383700 allocations: 33.11 MiB)`

Given the slight speed-up and that it will make the code cleaner, it might be worth getting rid of the TwoLayerParams, but there could be other reasons to keep it. Let me know what you think!

glwagner:
Fewer models is better: more people using the same code, less to maintain.

navidcy:
Let's remove it!

navidcy:
@mpudig thanks a lot for doing the benchmarks! The PR cleans up the module a lot!

I'll merge when tests pass unless there is an objection. 🚀

Issue #374, opened 2024-12-06
Add finite depth parameter to SQG model (`surfaceqg.jl`)


mncrowe:
I suggest adding a new depth parameter, say `H`, which can either be $\infty$, giving the current model, or finite.

In the finite case, the only change is that the Fourier space operator linking `b` and `ψ` changes from

$$ \hat{b} = |\textbf{k}|\hat{\psi}, $$

to

$$ \hat{b} = |\textbf{k}| \tanh(|\textbf{k}| H)\hat{\psi}. $$

From looking at the code, I think this just means changing the nonlinear function `calcN_advection!`, though some of the diagnostics might also need minor changes depending on how they're calculated.

Happy to help implement this.

glwagner:
Nice! It might make sense to build this feature as an option to precompute this function. That would be more flexible and also may save computational cost since `tanh` are relatively expensive.

mncrowe:
I've made a working version of this [here](https://github.com/mncrowe/GeophysicalFlows.jl). A few notes:

- This probably isn't how you want to handle types in `Params`. I'm guessing the way to go is similar to `SingleLayerQG` and have different structures for the finite and infinite parameter cases. I've also fudged the array type slightly in a way that isn't consistent with what you've got. I can sort this out if you let me how how you want to proceed with this.
- I haven't implemented any testing or examples currently.
- There doesn't seem to be any speed difference between the finite and infinite `H` cases with this implementation.
- No pull request currently, will do so once I'm a bit closer to a final version.

mncrowe:
PR [here](https://github.com/FourierFlows/GeophysicalFlows.jl/pull/375).

navidcy:
Can we combine the two definitions (finite and infinite depth) given that:
```Julia
julia> tanh(Inf)
1.0
```
?

What if there is a depth parameter with default value `Inf`?

navidcy:
Fair point about tanh being slow; let's keep it as is.

PR #375, opened 2025-01-21
Add finite depth functionality to `SurfaceQG`

mncrowe:
Addresses [issue 374](https://github.com/FourierFlows/GeophysicalFlows.jl/issues/374)


mncrowe:
I've added multiple dispatch to replace the if-else logic. CI failing for documentation, not sure why.

src/surfaceqg.jl
```diff
@@ -92,23 +98,32 @@ end
 # Parameters
 # ----------
 
+abstract type SurfaceQGParams <: AbstractParams end
+
 """
-    Params{T}(ν, nν, calcF!)
+    struct Params{T, Atrans}(H, ν, nν, calcF!, DirNeu) <: SurfaceQGParams
```
navidcy, line 104:
```suggestion
    struct Params{T, Atrans} <: SurfaceQGParams
```

navidcy (commented):
Suggestion:

At the moment there is no variable for stream function. If we introduce `ψh` in Vars then we can call `streamfunctionfromb!` within `calcN_advection!`. This way, by having multiple dispatch on `streamfunctionfromb!` we'll sort out the finite/infinite depth cases without having to almost-duplicate `calcN_advection!`. What do you think?

navidcy:
@mncrowe I tried to simplify. What do you think?

src/surfaceqg.jl
```diff
@@ -4,6 +4,7 @@ export
   Problem,
   set_b!,
   updatevars!,
+  streamfunctionfromb!,
```
navidcy, line 7: In the end this is not used, right? Delete?

src/surfaceqg.jl
```diff
@@ -335,33 +351,48 @@ Set the solution `sol` as the transform of `b` and update all variables.
...
+    streamfunctionfromb!(ψ, bh, params, grid)
+    streamfunctionfromb!(ψ, prob)
+
+Compute the streamfunction `ψ` from the buoyancy `bh` in Fourier space.
+(Note that `bh = prob.sol`.)
+"""
+function streamfunctionfromb!(ψ, bh, params, grid)
+  ldiv!(ψ, grid.rfftplan, @. params.ψhfrombh * bh)
   return nothing
 end
```
navidcy, line 370: Delete since it's not use?

navidcy:
This is ready if we add some tests.

Also if you wanna include the new diagnostic perhaps also a test for that?

src/surfaceqg.jl
mncrowe: It's not used. However, it might still be useful as the streamfunction is not available in `vars` and often it's very useful to know. Removing it would be ok as we have the inversion relation saved, though it's perhaps more user-friendly to have a function that calculates it rather than expecting users to do the Fourier transforms themselves?

src/surfaceqg.jl
mncrowe: same as above

mncrowe:
> This is ready if we add some tests.
> 
> Also if you wanna include the new diagnostic perhaps also a test for that?

Will add the diagnostic next week and think about some tests for that. A simple way to test the finite depth bit would be to just create a loop on `H in (Inf, 1)` for those tests where the value of `H` leads to different behaviour right?

src/surfaceqg.jl
glwagner: Note that "exporting" is really about choosing a stable user-API (not necessarily defining all functions that users need). Its ok to require users to qualify their names or write `using GeophysicalFlows: streamfunctionfromb!`. Not expressing an opinion about exporting or not here, just want to make sure that the goals are clear.

mncrowe: Yeah, this reply was intended to answer this and the question below in one go so I didn't separate removing the function and removing the export. I'll leave the choice of exporting vs not to you but I think including it (or perhaps a non-! version) would be useful.

glwagner: ah I was confused about @navidcy's suggestion

navidcy: ok, sorry if I created confusion.
I think I understand now the point of the method

I agree with @glwagner, we can include the method (if we also test it) but we don't necessarily need to export it.

src/surfaceqg.jl
navidcy: Since the method anyway allocates memory, wouldn't it be simpler if it was something like:

```Julia
function get_streamfunction(prob)
  ψh = @. prob.params.ψhfrombh * prob.sol
  return irfft(ψh, prob.grid.nx)
end
```

mncrowe: Yes, looks good, this is what I was getting at with 'non-! version' above.

navidcy: We use ldiv! and mul!  to avoid allocations when already the arrays we’ll put the values in are pre-existing.

navidcy:
We should include some finite-depth info at the docs module page:

https://github.com/mncrowe/GeophysicalFlows.jl/blob/main/docs/src/modules/surfaceqg.md

mncrowe:
> We should include some finite-depth info at the docs module page

I agree, I'll type out some stuff when I get a chance

mncrowe:
I've made a few additions:

1. Added some details on the finite depth case to the SQG documentation.
2. Verified that the `total_energy` diagnostic works as intended for SQG and added it to the list of exported functions.
3. Modified some of the existing SQG tests to apply to the finite depth case and added a loop on `H` to `runtests.jl`

The CI is now passing. What else do we need to do before this can be merged?

navidcy:
@mncrowe if you think it's ready I'll have a look!

navidcy:
Is there any test for finite H being added?

mncrowe:
> Is there any test for finite H being added?

Yes, I modified the energy & buoyancy variance test to include finite `H` and also test the new diagnostic (and also fixed a bug where it was passing by coincidence as the calculation was wrong). I also added finite `H` functionality to the advection and params constructor tests. These modified tests are included in a loop over `H in (1.0, Inf)` in `runtests.jl`.

https://github.com/mncrowe/GeophysicalFlows.jl/blob/38a63a5bca112fb7700e21bd384d887839d4e1f2/test/runtests.jl#L119C5-L124C8

https://github.com/mncrowe/GeophysicalFlows.jl/blob/38a63a5bca112fb7700e21bd384d887839d4e1f2/test/test_surfaceqg.jl#L35C1-L125C4

docs/src/lib/functions.md
```diff
@@ -94,6 +94,7 @@ GeophysicalFlows.MultiLayerQG.pv_streamfunction_kernel!
 
 ```@docs
 GeophysicalFlows.SurfaceQG.Problem
+GeophysicalFlows.get_streamfunction
```
navidcy, line 97:
```suggestion
GeophysicalFlows.SurfaceQG.get_streamfunction
```

docs/src/modules/surfaceqg.md
```diff
@@ -26,15 +26,21 @@ is equal and opposite to the vertical stretching of the buoyancy layers,
 \underbrace{\left(\partial_x^2 + \partial_y^2 \right) \psi}_{\textrm{relative vorticity}} + \underbrace{\partial_z^2 \psi}_{\textrm{stretching term}} = 0 ,
 ```
 
-with the boundary conditions ``b_s = - \partial_z \psi|_{z=0}`` and ``\psi \rightarrow 0`` as ``z \rightarrow -\infty``. (We take here the oceanographic convention: ``z \le 0``.)
+with the boundary conditions ``b_s = - \partial_z \psi|_{z=0}`` and ``\partial_z \psi|_{z=-H} = 0`` where ``H`` is the layer depth. In the case of infinite depth, ``H \rightarrow \infty``, the bottom boundary condition becomes ``\psi \rightarrow 0`` as ``z \rightarrow -\infty``. (We take here the oceanographic convention: ``z \le 0``).
```
navidcy, line 29:
```suggestion
with the boundary conditions ``b_s = - \partial_z \psi|_{z=0}`` and ``\partial_z \psi|_{z=-H} = 0``, where ``H`` is the layer depth. In the case of infinite depth, ``H \rightarrow \infty``, the bottom boundary condition becomes ``\psi \rightarrow 0`` as ``z \rightarrow -\infty``. (We take here the oceanographic convention: ``z \le 0``).
```

docs/src/modules/surfaceqg.md
```diff
@@ -26,15 +26,21 @@ is equal and opposite to the vertical stretching of the buoyancy layers,
...
+```
+
+for finite ``H`` and
+
 ```math
 \widehat{\psi}(k_x, k_y, z, t) = - \frac{\widehat{b_s}}{|𝐤|} \, e^{|𝐤|z} ,
 ```
 
-where ``|𝐤| = \sqrt{k_x^2 + k_y^2}`` is the total horizontal wavenumber.
+for the case ``H \rightarrow \infty``. Here ``|𝐤| = \sqrt{k_x^2 + k_y^2}`` is the total horizontal wavenumber.
```
navidcy, line 43:
```suggestion
for a fluid of infinite depth (``H \rightarrow \infty``). Here, ``|𝐤| = \sqrt{k_x^2 + k_y^2}`` is the total horizontal wavenumber.
```

src/surfaceqg.jl
```diff
@@ -49,6 +52,7 @@ Keyword arguments
   - `ny`: Number of grid points in ``y``-domain.
   - `Lx`: Extent of the ``x``-domain.
   - `Ly`: Extent of the ``y``-domain.
+  - `H`: Layer depth, set Inf for standard SQG.
```
navidcy, line 55:
```suggestion
  - `H`: Layer depth; `Inf` gives standard infinite-depth SQG dynamics.
```

navidcy (approved):
Thanks @mncrowe!

navidcy (commented):
@mncrowe some questions:

here:
https://github.com/mncrowe/GeophysicalFlows.jl/blob/639f7460e6f963e3ce17496671a1475c9b1787a8/src/surfaceqg.jl#L379-L428

1. I removed the `_3` for the gradient. I'd prefer `\nabla_h` for horizontal and just `\nabla` for the 3D. Could you check that I didn't introduce a mistake?

2. Why the total energy has an extra integral over `𝖽z`? How does this work out dimensionally. Is this a mistake on intentional?

navidcy:
Generally this is very nice contribution.
I'm only hesitant with the notation + expression in the energy docstrings. Also a bit worried that we have overloaded $\boldsymbol{k}$ to mean both 2D and 3D in different places (but that possibly happened even before this PR?)

mncrowe:
> Generally this is very nice contribution!
> 
> I'm only hesitant with the notation + expression in the energy docstrings. Also a bit worried that we have overloaded k to mean both 2D and 3D in different places (but that possibly happened even before this PR?)

In the `total_energy` docstring, ``𝐤`` refers to the 2D wave-vector, having done some divergence theorem to get rid of the 3D dependence. Is there an 3D wave-vector elsewhere? I've tried to use ``𝛁_h`` as the 2D gradient and ``𝛁`` as the 3D gradient so I *think* everything is consistent (at least in SQG, not sure how ``𝛁`` is used elsewhere). Could add subscript 3 to denote these quantities?

navidcy:
OK gotcha. I'll have another check.

Questions:

1. What's this dz here for?

https://github.com/mncrowe/GeophysicalFlows.jl/blob/17d3443f4c5c2e50401f6626a3100d2f0981e6a1/src/surfaceqg.jl#L423

2. Why isn't `total_energy` the some of the `buoyancy_variance` and `kinetic_energy`?

navidcy:
oh.. regarding `dz` it's what you explain at https://github.com/FourierFlows/GeophysicalFlows.jl/issues/376#issue-2804476736, right?

mncrowe:
> oh.. regarding `dz` it's what you explain at [#376 (comment)](https://github.com/FourierFlows/GeophysicalFlows.jl/issues/376#issue-2804476736), right?

Yes, essentially you convert a 3D volume integral into two surface integrals over the top and bottom. The bottom one vanishes as ``b = 0`` here so we're just left with a 2D result which can be put into (2D) Fourier space using a parseval identity.

mncrowe:
> 2. Why isn't `total_energy` the some of the `buoyancy_variance` and `kinetic_energy`?

Total energy is the sum of KE and PE *over the 3D volume*. Whereas the KE and buoyancy variance as you originally defined them are only the surface integrated value of these.

Because we have an evolution equation for $b$ on the surface, we can multiply by $b$ to show that the surface integrated value of $b^2$ is conserved. *Coincidentally*, this equals the surface integrated KE in the infinite depth case. There's no reason why this should be conserved from a physical perspective (I think), it's just that it happens to equal the buoyancy variance which is conserved. In the finite depth case, surface KE isn't conserved as it doesn't equal the buoyancy variance.

A second, independent, conserved quantity is the total energy which again we can show is conserved using the full 3D QG equation. And in the SQG case, this can conveniently be reduced to just an integral over the top surface. In fact it's an integral of $b \times \psi$ so it's not the same as buoyancy variance ($b \times b$) or surface KE ($|\nabla_h \psi|^2$).

navidcy (approved):

PR #397, opened 2025-10-07
Update CUDA and JLD2 compatibility versions

navidcy:


navidcy:
The CUDA upper bound to 5.8.3 is required because 5.8.4 and 5.9.0 suffer from https://github.com/JuliaGPU/CUDA.jl/issues/2903. When 5.8.5 or 5.9.1 is tagged the problem will disappear as https://github.com/JuliaGPU/CUDA.jl/pull/2907 will be included.

PR #412, opened 2026-02-26
Add simple 3D QG Model

mncrowe:
This PR adds a simple 3D QG model, based on the multilayer model. The model consists of PV advection in the interior, and buoyancy advection on the boundaries. Currently very much a work in progress.

I've chosen to use spectral collocation for the vertical grid and stored `z` in the `Params` object rather than the grid. Spectral collocation means that the PV inversion operator is dense, but only a small number of layers are needed in the vertical direction. The top and bottom rows of the system are replaced with the top and bottom buoyancy advection equations.

Many useful things aren't currently implemented (e.g. energy, stochastic stuff, forcing functions ...).

### DIFF
--- a/src/threedqg.jl
+++ b/src/threedqg.jl
@@ -1,0 +1,571 @@
+module ThreeDQG
+
+export
+  Problem
+
+  
+
+using
+  FFTW,
+  CUDA,
+  LinearAlgebra,
+  StaticArrays,
+  #Reexport,
+  #DocStringExtensions,
+  KernelAbstractions
+
+#@reexport using FourierFlows
+using FourierFlows
+
+using FourierFlows: parsevalsum, parsevalsum2, superzeros, plan_flows_rfft, CPU, GPU
+using KernelAbstractions.Extras.LoopInfo: @unroll
+
+#nothingfunction(args...) = nothing
+
+
+
+
+
+"""
+    SpecMatrix(x, w, w′)
+
+
+
+"""
+function SpecMatrix(x::AbstractVector, w::AbstractVector, w′::AbstractVector)
+
+    N = length(x)
+    D = x' .- x
+
+    D₁ = D
+    D₁[D .== 0] .= 1
+
+    A = ones(N, 1) * (w' .* prod(D₁; dims = 1))
+
+    M = A' ./ (A .* D')
+    M[1:(N+1):N^2] .= sum(1 ./ D₁; dims = 1)' .- 1 + (w′ ./ w)
+
+    return M
+
+end
+
+"""
+    SpecMatrix(x)
+
+
+
+"""
+SpecMatrix(x) = SpecMatrix(x, ones(length(x)), zeros(length(x)))
+
+
+"""
+    GridChebyshev(N, L)
+
+Creates a grid of Chebyshev points of the second kind and a spectral collocation
+differentiation matrix.
+
+# Arguments:
+ - `N`: number of gridpoints
+ - `L`: vector of endpoints, ``x ∈ [L₁, L₂]``
+
+"""
+function GridChebyshev(N::Int, L::Vector)
+
+    x = (L[1] + L[2]) / 2 .- (L[2] - L[1]) / 2 * cos.((0:(N-1))*π/(N-1))
+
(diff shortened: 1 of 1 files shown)


src/threedqg.jl
```diff
@@ -0,0 +1,571 @@
...
+    D₁ = D
+    D₁[D .== 0] .= 1
+
+    A = ones(N, 1) * (w' .* prod(D₁; dims = 1))
+
+    M = A' ./ (A .* D')
+    M[1:(N+1):N^2] .= sum(1 ./ D₁; dims = 1)' .- 1 + (w′ ./ w)
+
+    return M
+
```
glwagner, line 49:
```suggestion
```

src/threedqg.jl
```diff
@@ -0,0 +1,571 @@
...
+"""
+function GridChebyshev(N::Int, L::Vector)
+
+    x = (L[1] + L[2]) / 2 .- (L[2] - L[1]) / 2 * cos.((0:(N-1))*π/(N-1))
+
+    M = SpecMatrix(x)
+
+    return x, M
+
```
glwagner, line 79:
```suggestion
```

src/threedqg.jl
```diff
@@ -0,0 +1,571 @@
...
+    else
+        x = [0; p]
+        w = [1; exp.(-p/2)]
+        w′ = -1 / 2 * w
+    end
+
+    M = SpecMatrix(x, w, w′)
+
+    return x, M
+
```
glwagner, line 104:
```suggestion
```

src/threedqg.jl
```diff
@@ -0,0 +1,571 @@
...
+Creates a grid of Chebyshev points of the second kind and a spectral collocation
+differentiation matrix.
+
+# Arguments:
+ - `N`: number of gridpoints
+ - `L`: vector of endpoints, ``x ∈ [L₁, L₂]``
+
+"""
+function GridChebyshev(N::Int, L::Vector)
```
glwagner, line 72: This is very interesting! Should this somehow be added to `FourierFlows`? Wondering if there should be one function for

```julia
chebyshev_nodes
```

and then the `spectral_matrix` can be built downstream, what do you think?

glwagner: I see, below the `spectral_matrix` (is that what is meant by SpecMatrix?) has additional arguments ...

maybe we need `chebyshev_nodes` and `laguerre_nodes` in FourierFlows, and then here we need `chebyshev_matrix` and `laguerre_matrix`?

mncrowe: > This is very interesting! Should this somehow be added to `FourierFlows`? Wondering if there should be one function for
> 
> ```julia
> chebyshev_nodes
> ```
> 
> and then the `spectral_matrix` can be built downstream, what do you think?

Yes, certainly. Currently what I have is just copied from my [coastal trapped wave](https://github.com/mncrowe/CTWSolve.jl) code which is structured very differently. I think it looks a bit untidy to have `z` in Params rather than the grid so some alternate grid structure (via `FourierFlows`) is maybe the way to go?

glwagner: I think the cleanest way is to generalize ThreeDGrid: https://github.com/FourierFlows/FourierFlows.jl/blob/c04ed88be41c062d423e69f4a56ee76dc9245c87/src/domains.jl#L232

somehow you have to abstract the z-coordinate part, so that you can have `CheyshevVerticalCoordinate`, etc?

mncrowe: Is it worth making a grid type that's a subtype of the `ThreeDGrid` as the fields I'd need aren't consistent with `ThreeDGrid` in it's current form? Then modify the function which creates this grid with an extra keyword `z_type = :fourer, :chebyshev, :laguerre` which defaults to `:fourier`?. Also, I think the `spectral_matrix` should be built when the grid is constructed, as it's the analogue of the `fftplan` for the Fourier grid. Will draft something when I get a few minutes.

glwagner: well, I was suggesting that you _change_ `ThreeDGrid` so that it is consisent with what you're trying to do. This is what I mean by "generalization".

I understand if its too much work though.

